PR #4, opened 2015-03-30
Language fixes in readme

michielbdejong:
* third person singular gets -s for  most verbs in English
* adjectives don't get an -s, even when they apply to plural objects
* software is not countable (cannot be used plural). Same for information.
* 'each' goes with singular, not plural
* some word order fixes
* some collocation fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,18 @@
 Clouder
 -------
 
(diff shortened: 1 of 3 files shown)


YannickB:
Thank you Michiel, I have a lot of writing to do currently and this kind of review are very welcome. Congratulations for your first :)

PR #34, opened 2015-06-23
postgresql specific version installation

MarcEtienneDery:
Installing a specific version of Postgresql to avoid any problem in the future.


YannickB:
+1 good idea, thank you

PR #37, opened 2015-07-01
implement purging of image from the private registry

### LINKED ISSUE #29: Improvement: purge image
https://github.com/clouder-community/clouder/blob/8.0/clouder/clouder_image.py#L282 
You could use the docker registry api, http://docs.docker.com/registry/spec/api/#deleting-an-image

MarcEtienneDery:
fixes #29

### DIFF
--- a/clouder/clouder_image.py
+++ b/clouder/clouder_image.py
@@ -281,9 +281,9 @@ class ClouderImageVersion(models.Model):
         """
         Delete an image from the private registry.
         """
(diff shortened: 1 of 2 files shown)


YannickB:
Thank you, I'll test it as soon as I can.

MarcEtienneDery:
I can't use the fullpath_locahost since I'll be inside the registry container and I need to use the localpost instead of the hostport.

YannickB:
You're right. Ok looks good for me, thank you for your contribution.

PR #38, opened 2015-07-02
Update clouder_container.py

ErrafayM:
Container always start/restart on boot


YannickB:
Nice catch, thank you

PR #59, opened 2015-11-12
Update clouder_template_backup_data.xml

ErrafayM:
I had this problem while building the backup image

stderr : The command '/bin/sh -c git clone git://github.com/bup/bup /opt/bup' returned a non-zero code: 128

I changed the link with the one in bup/bup git and started working just fine. I dont know if i'm the only one who had this issue. can you check this out Yannick?


YannickB:
Hum I know I often have problem with git clone git:// if I am not authenticated, I agree using https:// is safer

PR #81, opened 2015-12-18
Making BIND generic

pasgou:
I Propose the base of a modification for the BIND template.

What i don't know how to do : making that the user entrie is correct and acceptable by bind (is it an IP address or "none"?) and making the code is sure to make impossible the user to break down the servers.

### DIFF
--- a/clouder_template_bind/clouder_template_bind.py
+++ b/clouder_template_bind/clouder_template_bind.py
@@ -58,7 +58,7 @@ class ClouderDomain(models.Model):
             self.execute(ssh, [
                 "echo 'zone \"" + self.name + "\" {' >> /etc/bind/named.conf"])
             self.execute(ssh, ['echo "type master;" >> /etc/bind/named.conf'])
-            self.execute(ssh, ['echo "allow-transfer { '+ self.options['Slave_IP']['value'] +';};" '
+            self.execute(ssh, ['echo "allow-transfer { '+ self.options['slave_ip']['value'] +';};" '
                                '>> /etc/bind/named.conf'])
             self.execute(ssh, ["echo 'file \"/etc/bind/db." +
                                self.name + "\";' >> /etc/bind/named.conf"])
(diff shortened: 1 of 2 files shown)


YannickB:
Hi,

It looks ok for me with only two little correction : 
-you should use only lesser case for the name of the option, so "Slave_IP" should be renamed "slave_ip".
-I think allow_transfor is not a required configuration, so if the option isn't filled we should not set this line. Otherwise if the user do not fill the option, the allow_transfer will be empty and may cause an error.

Thank you!

pasgou:
Ok

Could you correct the second point as I don’t know how to.

In the documentation i have read ( http://www.zytrax.com/books/dns/ch6/ <http://www.zytrax.com/books/dns/ch6/> ) , the allow_transfer option can be at « none », or be an IP, or not to be.
what I would like to know, for my knowledge, is why we could not have an option to configure or a DNS master or a DNS for sue-domain delegation.

Cordialement

Pascal GOUHIER

pasgou:
Lesser case corrected.
I dont know how to correct the second part.

pasgou:
New comment for the generic bind container.

In the Clouder concept, all services of the company are driven from Clouder : ERP (Odoo), Blog (Drupal or Wordpress), Supervision (Shrinken), Mail (Postfix) and even DNS (Bind).

But what if we have some services that aren't driven from Clouder? Example : we have Google for Business with our Domain Name, a payroll server on a TSE or other services that can't be managed from Clouder. What could be great is : 
- Or to be able to manage the bind service from clouder without the pre-required Application entry
- Or to be able to manage only part of the zones
But what if the bind service must be a "master".

What are your advice? Or have you a trick?

YannickB:
Hello, 

The bind container is quite basic currently, it only add some basic configuration and add the A record when you create a new base. You are perfectly free to go in the bind container and make some change directly on it.

Regarding Google for Business, I think we should improve the dns template to be able to have a "fake" bind container with in the option the Google for business credentials. If this container is selected in the domain link, it will connect to google api to add the A record instead of adding them in the bind container.

pasgou:
Thank you for your answer.
I will study this way of doing things.

PR #83, opened 2016-01-05
Fixing a bug in clouder_template_postfix

nicolas-petit:
Added a space at the end of a file to make it PEP-8 complient.
Added conditions to make sure SMTP options are set only when all required
    options are filled.


pasgou:
It s a pleasure to read code cause I try to understand python coding and I thank you for your attention.
I take note about this remark.

YannickB:
Thank for the contribution, merge accepted

PR #84, opened 2016-01-06
Fixed a very probable bug on container's test_connect function.

nicolas-petit:
The function never read the ssh_port option on containers, and
    called port 0 instead.


YannickB:
We discussed about this during a phonecall, not a bug

PR #87, opened 2016-01-18
Sanitize o2m bug

YannickB:


clouder/base.py
```diff
@@ -248,54 +248,162 @@ def onchange_application_id_vals(self, vals):
...
-            ).iteritems():
-                if option.source.type == 'base' and option.source.auto:
-                    if option.source.app_code and \
-                            option.source.app_code != application.code:
-                        continue
-                    options.append((0, 0, {
-                        'name': option.source.id,
-                        'value': getattr(option, 'value', False) or option.source.get_default}))
+            # Getting sources for new options
+            option_sources = {x.id: x for x in application.type_id.option_ids}
```
YannickB: Only options with type = 'base' and auto = True shall be added in option_sources. Check line 254 in old file.

clouder/base.py
```diff
@@ -248,54 +248,162 @@ def onchange_application_id_vals(self, vals):
                     or self.email_sysadmin
 
             options = []
-            for key, option in self.sanitize_o2m(
-                    'option', vals, application.type_id.option_ids, True
-            ).iteritems():
-                if option.source.type == 'base' and option.source.auto:
```
YannickB: This code is important and shall not disappear. You have an equivalent in all others one2many fields and you also need to correct them.

clouder/base.py
```diff
@@ -248,54 +248,162 @@ def onchange_application_id_vals(self, vals):
...
-                            option.source.app_code != application.code:
-                        continue
-                    options.append((0, 0, {
-                        'name': option.source.id,
-                        'value': getattr(option, 'value', False) or option.source.get_default}))
+            # Getting sources for new options
+            option_sources = {x.id: x for x in application.type_id.option_ids}
+            sources_to_add = option_sources.keys()
+            # Checking old options
+            if 'option_ids' in vals:
```
YannickB: I think this line is too much, because we want to set the automatic options when we create a new base, even if we didn't set manually some lines in option_ids

PR #101, opened 2016-06-07
Web create

nicolas-petit:
The plugin is ready to be updated with the instance creation function.


YannickB:
Good job, thank you

-Rename the module into clouder_website
-Rename the TestOdoo variable we can see everywhere into something more appropriate
-Add the host adress in the js configuration. Plugin.js should be served by the odoo server (or wsgi)
-Translate french comment into english

clouder/base.py
```diff
@@ -248,8 +248,8 @@ def onchange_application_id_vals(self, vals):
                 vals['application_id'])
 
             if 'admin_name' not in vals or not vals['admin_name']:
-                vals['admin_name'] = application.admin_name
-
+                vals['admin_name'] = application.admin_name and application.admin_name or \
```
YannickB: It's best to have the real email as login name, especially for Odoo. Which problem did you had to had these replaces?

clouder_template_odoo/template.xml
```diff
@@ -532,6 +532,12 @@
             <field name="name">/opt/odoo/files</field>
             <field name="nosave" eval="True"/>
         </record>
+        <record id="image_odoo_clouder_exec_port_js" model="clouder.image.port">
```
YannickB: We can remove it now

clouder_website/clouder_website.py
```diff
@@ -0,0 +1,915 @@
...
+class ClouderApplication(models.Model):
+    """
+    Adds information for web-creation on applications
+    """
+
+    _inherit = 'clouder.application'
+
+    web_create_type = fields.Selection(
+        [
+            ('disabled', 'Disabled'),
```
YannickB: Maybe a required field since we have the option disabled?

clouder/base.py
nicolas-petit: Judging by the change made and the commit number itselt (https://github.com/clouder-community/clouder/pull/101/commits/6cb052268e11ea388adc8567b12f47e96373cffa) I'd say the problem was:
- when admin_name wasn't present, the code tried to put in the email.
- email-specific characters were rejected from the 'admin_email' column, and the creation failed.


See the check for admin_email here:
 https://github.com/nicolas-petit/clouder/blob/6cb052268e11ea388adc8567b12f47e96373cffa/clouder/base.py#L213

clouder_template_odoo/template.xml
nicolas-petit: Removed the record from my local branch.
Scheduled for next commit.

clouder_website/clouder_website.py
nicolas-petit: A good point.
I've added it to my local changes.
Scheduled for next commit.

clouder/base.py
YannickB: Damn you're right, I'm not sure why I added theses check. Anyway if we use the email I want the real one, I don't want any replace() here. Please correct the check to allow it, we'll see if any problems arise later.

nicolas-petit: I'll replace the regexp with the one from the emails and restore the line.
Staged for next commit.

Issue #119, opened 2016-08-23
Continous Deployment


YannickB:
Hello everyone,

I'd like to raise an important issue regarding the next step of Clouder. Since last week we were working on integrating Gitlab in order to manage the versionning of the projects.

Fact is, I completely underestimated Gitlab. This is far more than a versionning and issue management tool now, with Gitlab CI it became a complete solution to manage the Continuous Integration and Continuous Deployment workflow, and I think it is a real opportunity to bring Clouder to the next and state of the art level.

Today we have finished the gitlab template and we already made a first work on the continous integration for the Odoo template. Of course we used the travis scripts of the OCA, there is still a lot of work to really have private continuous integration for Odoo and we will probably need to adapt some of the OCA scripts to make them work with Gitlab, but at the very least they are executed so we're already quite far in the continous integration feature.

More important is the continous deployment issue because it may fundamentally change the way Clouder work.
I see three types of applications : 
- The first are the applications without specific development and which are stable enough to be automatically updated. I'm thinking of the registry, backup, bind, shinken, postfix, postgres, mysql etc... Most of the subcomponents.
- The second are applications without specific developments but which contains some risks during update because it need database update. For example mautic, piwik or gitlab itself. We may want automatic update for minor version but keep manual update for major version to be sure technical team are ready to handle problem.
- The third are applications with specific developments, like Odoo, Wordpress, Drupal, and web applications in general. For them we need a real continuous deployment workflow, hence gitlab for versionning, continous integration for testing, and automatic deployment after successful test or manual validation.

Let's remind that in Clouder most of the applications, and the Odoo one in particular, use three containers : 
- Data, where data related to each bases and configuration are stored, for easier backup. In Odoo case, configuration files, filestore, etc...
- Files, which contains the files of the applications and is automatically build with the last version when we build the images. If we want to update the application, we just change the Files container and update the bases.
- Exec, which contains the binary and is linked to Data and Files. It's the container which serve the data and is being exposed, it need to be regularly updated to have security patchs. This container can be duplicated for scalability, while still being connected to the same data.
We can also have some other containers, like SSH to give access to the customers, but they are the three important ones. In the end, all applications (including subcomponent like bind/shinken/postgres etc...) will have a least Data and Exec in their templates.

Today Clouder need a lot of manual intervention. It's a great tool for initial deployment (and so it's perfect to integrate behind a website form) but keeping the containers updated is a complex workflow : 
- You have to go in the concerned images and rebuild them
[...]

PR #120, opened 2016-08-24
Gitlab CI and Registry refactoring

YannickB:
With this commit, the Odoo build and release in registry through Gitlab is now working. Next step is the deployment itself.

Theses commit break current installation so they'll be release in the future in a new version. Until them, we will use master branch


YannickB:
#119

PR #133, opened 2016-10-12
[WIP][FIX] Fix build errors for Clouder 0.9.0

lasley:
This is a work in progress. 

@YannickB would you mind verifying that I inferred these fixes correctly? It's going to be pretty much the same thing across all of the templates, so I want to make sure I'm doing it right before moving further.

High level of the inferences represented in this PR:
* `clouder.image.current_version` is no longer a static char, but instead a relation to `clouder.image.version`, in which the `name` field should be the string that was previously defined
* `clouder.image.privileged` logic has been removed with no replacement
* `clouder.application.*_saverepo_*` logic has been removed with no replacement
* `clouder.image.volume.from_code` logic has been removed with no replacement
* `clouder.application.link.service` logic has been remved with no replacement
* `clouder.application.parent_id` was removed erroneously & needed to be added back (child_ids m2m relation still had parent_id set as the relation col). 
 * Column naming of `parent_id` needed to be updated to conform to m2m standard of `_ids` prefix on multi relations.
* Magento is dependent on Piwik

### DIFF
--- a/clouder/application.py
+++ b/clouder/application.py
@@ -188,8 +188,19 @@ class ClouderApplication(models.Model):
     required = fields.Boolean('Required?')
     sequence = fields.Integer('Sequence')
     child_ids = fields.Many2many(
-        'clouder.application', 'clouder_application_parent_child_rel',
-        'parent_id', 'child_id', 'Childs')
+        'clouder.application',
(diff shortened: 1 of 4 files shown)


lasley:
Closing, these are fixed for the most part outside of the warnings.

PR #142, opened 2016-10-14
POC Route53

YannickB:


lasley:
cc @laslabs

YannickB:
Ok travis should be green now. @lasley what do you prefer, I directly merge the POC in master or you contribute on my branch ? Personally I prefer merging without waiting, I had to add the feature with the dummy container and I don't mind keeping the route53 files empty for some time.

lasley:
Assuming it's functional, I prefer the quick merge seeing as we're working on an incubator. Production would be a bit different, but we've already made it clear that this is a dev head & maintaining PRs sucks.

YannickB:
Way to go then

PR #144, opened 2016-10-16
[IMP] clouder: Path and Environment Handling

lasley:
[IMP] clouder: Centralize and improve path/str handling  …
* Centralize key and clouder tmp dirs via `ClouderModel` method
* Centralize backup directory definitions in `ClouderModel`
* Harden path handling by using os.path.join
* Prefer string interpolation vs concat
* Fix use of `self.raise_error` for models that aren't inherited from `clouder.model`
* Prefer `ClouderError` in models not inherited from `clouder.model`
* Update model attribute declaration orders
* Use lambdas for defaults in `clouder.container`
* Environment prefix is required
* Centralize commit methods and use isolated environment
* Change version to `9.0.2.0.0` to match OCA style for 9, while keeping w/ semantic increase

### DIFF
--- a/clouder/__openerp__.py
+++ b/clouder/__openerp__.py
@@ -22,7 +22,7 @@
 
 {
     'name': 'Clouder',
-    'version': '10.0.10.0.0',
+    'version': '9.0.2.0.0',
     'category': 'Clouder',
     'depends': ['base'],
     'author': 'Yannick Buron (Clouder), LasLabs',
--- a/clouder/__openerp__.py
+++ b/clouder/__openerp__.py
@@ -22,7 +22,7 @@
 
 {
     'name': 'Clouder',
-    'version': '9.0.2.0.0',
+    'version': '9.0.10.0.0',
     'category': 'Clouder',
     'depends': ['base'],
     'author': 'Yannick Buron (Clouder), LasLabs',
--- a/clouder/container.py
+++ b/clouder/container.py
@@ -23,6 +23,7 @@
 from openerp import models, fields, api
 from openerp import modules
 
+import os.path
 import socket
 import re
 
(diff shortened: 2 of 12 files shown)


clouder/save.py
```diff
@@ -266,142 +266,217 @@ def deploy(self):
         """
         Build the save and move it into the backup container.
         """
+
+        self.ensure_one()
```
lasley: I started doing this, then realized that there are a lot of methods expecting singletons without checking. I'll push that off into a model audit PR instead.

lasley:
Alright this seems to work in local & is ready to review assuming CI goes 🍏 

I updated the changelog in the PR description to contain a complete list of changes made (also documented in my commit msg)

Couple notes to bring out of the changelog:
* Directory handling is improved overall, with most string concatenation removed. Lots centralized
* Commits have been centralized into a `clouder.model` method providing a new env & commit in a context manager. This is a more transaction-safe approach while still maintaining original intent IMO
* The version should be `9.0.x.x.x` to indicate we're using v9. Connector isn't out for 10 yet, so probably not good to make that jump quite yet (also removes a lint error).

clouder/save.py
YannickB: Yeah I wasn't so used to the concept behind api.multi and api.one. I understood a lot last week when I removed api.one, I agree self.ensure_one shall be set in some functions in clouder.model.

clouder/__openerp__.py
```diff
@@ -22,7 +22,7 @@
 
 {
     'name': 'Clouder',
-    'version': '10.0.10.0.0',
+    'version': '9.0.2.0.0',
```
YannickB: Agree for the first number which is Odoo version number, but then it's Clouder version. I'd suggest either 9.0.10.0.0 since it's an incubator for next version, or keep 9.0.9.0.0 until we release it.

I wish I'd just set 0.10.0.0.0 or 0.9.0.0.0 but travis disagree.

clouder/environment.py
```diff
@@ -35,21 +37,21 @@ class ClouderEnvironment(models.Model):
...
+    _description = 'Clouder Environment'
+    _sql_constraints = [
+        ('prefix_uniq', 'unique(prefix)',
+         'Prefix must be unique!'),
+    ]
 
     name = fields.Char('Name', required=True)
     partner_id = fields.Many2one('res.partner', 'Partner', required=True)
-    prefix = fields.Char('Prefix', required=False)
+    prefix = fields.Char('Prefix', required=True)
```
YannickB: There is a reason why prefix is not required (it was before)

Access right are based on the environment model. Which means that when a customer will create a base from the subscription form, we will create an environment just for him so he can manage his bases.

If he using a public Odoo instance (containers which contains bases from several other customers) and don't create any containers, then we don't want him to bother trying to find an available prefix (since it has a unique constraint).

tldr : According to me prefix should not be required in database, but we should have an Odoo constraint which check that all environment linked to a container have a prefix. But that's a difficult question, I'm open for opinions.

lasley: Thanks for the elaboration, I knew there was something I was missing - particularly because of the `required=False` being explicitly defined. I'll switch this back for the moment - no immediate ideas pop out.

lasley:
@YannickB - I deployed a few containers, but yeah good call with the One Click deploy. I'm going to wait for @t3ddftw to give us a review as well - lots of changes here that I'd like more eyes on.

clouder/environment.py
YannickB: Yes usually when I set a required=False, it's because I need to remove a NOT NULL constraint on existing installation because it was set previously.

clouder/__openerp__.py
lasley: I'll change it back and we'll breach that gap when we get there then. We've got more important things to worry about at the moment 😉

clouder/environment.py
lasley: Typically it's best to do that via migrations so it's a bit more clear in terms of history, but whatever's clever!

YannickB:
Ok. During the oneclick process, note that the last step (the duplication of Odoo instance into a test install) was already broken before your changes. Everything else was working.

In general oneclick is the best way to check all the code, we can't use travis here because it's obviously kinda difficult to test an infrastructure installation with unit test (but I still hope we'll find a solution one day to automatise this part).

lasley:
I've got us covered on the testing of external services via unit tests, no problem ([connector-carepoint coverage](https://codecov.io/gh/laslabs/odoo-connector-carepoint), [connector-dns coverage](https://codecov.io/gh/laslabs/infrastructure-dns/branch/release%2F8.0%2Fconnector-dns)). 

Big undertaking though so I want to make sure we've got major changes done before tackling that.

YannickB:
That's good to know, agree we have other priorities but that's something I have in mind since quite a long time, good to see we have some clues.

clouder/container.py
```diff
@@ -189,26 +140,101 @@ def _inverse_public_key(self):
...
+        """
+        for server in self:
+            key_dir = os.path.join(server.home_directory, '.ssh', 'keys')
+            self.execute_local(['mkdir', '-p', key_dir])
+            key_file = os.path.join(key_dir, server.fulldomain)
+            key_file_pub = '%s.pub' % key_file
+            self.execute_write_file(
+                key_file_pub, server.public_key, operator='w',
+            )
+            self.execute_local(['chmod', '700', key_file_pub])
```
t3ddftw: @lasley,

Nice pick but the public key shouldn't need the x bit - 600 may be more appropriate.

Thanks!

clouder/container.py
```diff
@@ -189,26 +140,101 @@ def _inverse_public_key(self):
...
+                'mkdir', '-p',
+                os.path.join(server.home_directory, '.ssh', 'keys'),
+            ])
+            key_file = os.path.join(
+                server.home_directory, '.ssh', 'keys', name,
+            )
+            self.execute_write_file(
+                key_file, server.private_key, operator='w',
+            )
+            self.execute_local(['chmod', '700', key_file])
```
t3ddftw: @lasley,

Same as public key; should only need mode 600.

clouder/image.py
```diff
@@ -90,21 +93,22 @@ def computed_dockerfile(self):
...
-        dockerfile += \
-            '\nMAINTAINER ' + self.env['clouder.model'].email_sysadmin + '\n'
+        dockerfile += '\nMAINTAINER %s\n' % (
+            self.env['clouder.model'].email_sysadmin,
+        )
 
         dockerfile += self.dockerfile or ''
         volumes = ''
         for volume in self.volume_ids:
             volumes += volume.name + ' '
```
t3ddftw: May be out of scope but String concat is significantly slower than, say, joining a list of strings (as I'm sure you know).

https://waymoot.org/home/python_string/

lasley:
You're spammy @t3ddftw - please use the "review" feature instead of "single comment"

clouder/image.py
lasley: I am trying to nail all the concats as well, so yup in scope

That said, string concat has been improved drastically in newer Pythons. It's still susceptible to the n+1 issue of immutable joins in terms of memory overhead, the CPU time has been fixed for the most part.

clouder/container.py
```diff
@@ -49,113 +50,57 @@ class ClouderServer(models.Model):
 
     _name = 'clouder.server'
     _inherit = ['clouder.model']
+    _sql_constraints = [
+        ('name_uniq', 'unique(name, domain_id)',
+         'Name must be unique!'),
+        ('ip_uniq', 'unique(ip, ssh_port)',
+         'IP/SSH must be unique!'),
```
t3ddftw: I read the help before looking at the code and was confused - Perhaps 'IP/SSH Port must be unique!' would be a better way to word it? Or I'm just (derp)

clouder/model.py
```diff
@@ -626,9 +690,9 @@ def send(self, source, destination, ssh=False, username=False):
...
         tmp_dir = False
         if self != server:
-            tmp_dir = '/tmp/clouder/' + str(time.time())
+            tmp_dir = self.get_directory_clouder(time.time())
             server.execute(['mkdir', '-p', tmp_dir])
-            destination = tmp_dir + '/file'
+            destination = os.path.join(tmp_dir, 'file')
 
         sftp = ssh.open_sftp()
         self.log('send : ' + source + ' to ' + destination)
```
lasley: interpolate string

clouder/save.py
```diff
@@ -266,142 +266,217 @@ def deploy(self):
...
+
         self.log('Saving ' + self.name)
         self.log('Comment: ' + self.comment)
```
lasley: interpolate strings

clouder/container.py
```diff
@@ -339,9 +368,9 @@ def purge_dns_exec(self):
         self.control_dns = False
         if self.domain_id.dns_id:
             self.domain_id.dns_id.execute([
-                'sed', '-i',
-                '"/' + self.name + r'\sIN\sA/d"',
-                self.domain_id.configfile])
+                'sed', '-i', r'"/%s\sIN\sA/d"' % self.name,
```
lasley: @YannickB - The `"` in this are getting escaped when sending to the server. Is this intentional?

```
docker exec dev-salt-master-exec bash -c  "sed -i \"/containers\/dev-dfgdrgdrt/d\" /srv/pillar/top.sls "
```

I double checked and it's doing the same on 0.9.0

lasley:
Thanks for the reviews @YannickB & @t3ddftw - comments attended to. Haven't functionally reviewed fully yet, so still WIP.

Description of latest changes:
* Add dynamic values to Registry X509 cert
* Make generate_random_password somewhat more secure
* Use lambdas for password field defaults in order to avoid re-use of passwords
* Remove execution bit from key files
* Reverse container sql constraint for domain/name for efficiency
* Improve container sql constraint verbiage & names
* Change version to 9.0.10.0.0
* Prefix is not required on environment
* Remove more string concats

In hindsight, the top three I probably should have done in separate PRs.

clouder/container.py
YannickB: +1 600 is enough. Note that the fault is mine, he only copy/paste my code.

clouder/container.py
YannickB: Yes this is intentional, this is the argument of the sed command and `"` were already used for the docker exec argument. I guess this command is extracted from the log, `"` are automatically escaped in docker exec command.

YannickB:
Ok looks good enough for me, let me know when I can merge.

lasley:
FYI there's an issue with this that I have been resolving. Probably shouldn't have made so many changes at once - debugging has been a real pain. Too late now though 😆

lasley:
Alright this seems to work now.

I had to remove the transaction handling for the moment in order to minimize the scope of this ticket troubleshooting. I left the centralization of it in though, which will make for an easier refactor once I study the environment a bit more. I'm going to make an issue for that.

YannickB:
Ok, should I merge it now ?

lasley:
Yup we're good to merge 👍

YannickB:
And done, thanks for everything

PR #145, opened 2016-10-16
[IMP] Migrate to v10

lasley:
- Increase Odoo version to 10.0 in Travis File

Depends:
- [x] OCA/connector#211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ addons:
 
 env:
   global:
-  - VERSION="9.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
 
   matrix:
   - LINT_CHECK="1"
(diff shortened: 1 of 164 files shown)


lasley:
@YannickB - How important is it to keep async activated during incubator? I would like to deactivate it so that we can move up to 10.

Reason being - there are methods existing in the new model API that will make things much more efficient. I'd like to start knocking them out, because they'll be easier to back-port due to broken CI than to forward-port due to having to re-audit.

YannickB:
Now that we moved async out of the core it's completely optional and I don't think we'll use it before long. You can deactivate.

lasley:
Sweet! Admittedly I'm just getting tired of having to click the stupid arrow in the M2O fields. v10 you can click anywhere and it pains me to move five millimeters to the right

YannickB:
Ahah, well we have to move to v10 anyway.

lasley:
Alright this looks to work just by the simple upgrade regexes! I love it when that happens, although we'll probably find some errors here and there just due to lack of test coverage. 

You can do a functional test using [this Runbot](https://09679-145-2cc934.runbot.laslabs.io/web?db=09679-145-2cc934-all).

YannickB:
Great :), ready to merge then ?

lasley:
Yup!

Issue #147, opened 2016-10-24
[BUG][master] No matching record found for external id 'connector.model_queue_job'


lasley:
This is primarily a note to self, but anyone is welcome to fix 😉 

Pretty sure this just needs to be moved back into the async module?

To replicate, install `clouder` on a fresh database.

Branch `master`

```
2016-10-24 16:20:23,547 1040 ERROR clouder openerp.modules.registry: Failed to load registry
Traceback (most recent call last):
  File "/var/www/odoo/openerp/modules/registry.py", line 386, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/var/www/odoo/openerp/modules/loading.py", line 338, in load_modules
    loaded_modules, update_module)
  File "/var/www/odoo/openerp/modules/loading.py", line 237, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
  File "/var/www/odoo/openerp/modules/loading.py", line 156, in load_module_graph
    _load_data(cr, module_name, idref, mode, kind='data')
  File "/var/www/odoo/openerp/modules/loading.py", line 98, in _load_data
    tools.convert_file(cr, module_name, filename, idref, mode, noupdate, kind, report)
  File "/var/www/odoo/openerp/tools/convert.py", line 845, in convert_file
    convert_csv_import(cr, module, pathname, fp.read(), idref, mode, noupdate)
  File "/var/www/odoo/openerp/tools/convert.py", line 913, in convert_csv_import
    raise Exception(_('Module loading %s failed: file %s could not be processed:\n %s') % (module, fname, warning_msg))
Exception: Module loading clouder failed: file clouder/security/ir.model.access.csv could not be processed:
 Line 1 : No matching record found for external id 'connector.model_queue_job' in field 'Object'
2016-10-24 16:20:23,548 1040 ERROR clouder openerp.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/var/www/odoo/openerp/http.py", line 647, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/var/www/odoo/openerp/http.py", line 684, in dispatch
    result = self._call_function(**self.params)
  File "/var/www/odoo/openerp/http.py", line 320, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/var/www/odoo/openerp/service/model.py", line 118, in wrapper
[...]

PR #150, opened 2016-10-26
Update __openerp__.py - add connector dependency

haylahi:
minor fix - add connector dependency

### DIFF
--- a/clouder/__openerp__.py
+++ b/clouder/__openerp__.py
@@ -24,7 +24,7 @@
     'name': 'Clouder',
     'version': '9.0.10.0.0',
     'category': 'Clouder',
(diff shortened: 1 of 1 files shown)


lasley:
Fixes #147, caused by #143 

Inclusion of this causes the need for AGPL license in clouder core

YannickB:
Hello,

We separated the asynchronous feature (using the connector) in the clouder_asynchronous module to break the AGPL incompatbility. We don't want the core module to inherit connector, which issues did you had ?

lasley:
This was fixed in #161 by moving the permissions back to the async module. Everything works as intended now & this PR may be closed.

PR #151, opened 2016-10-27
[IMP][master] Split model files by name

lasley:
No functional changes here, just a step towards OCA compliance by splitting models by name & updating the copyright/license headers to shorter version.

[IMP] clouder: Split model files by name  …
- Isolate model files by model name

[FIX] clouder_template_mysql: Fix model import path

[FIX] clouder_template_gitlab: Fix model import path

**Edit**

Looks like my quick deploy passed too. I did get saddened when I realized no quick deploy multiple applications to the same server though 😦


YannickB:
Wow so much files now, I'm a little sad that we have to make one file for small object like port/volumes/etc.. but I guess we have no choice. Ok for me, can I merge ?

lasley:
Yup good to merge.

Technically we could combine the smaller models if we want - OCA guidelines do allow that. I was just on a role of separating them and wasn't paying an incredible amount of attention TBH. 

One thing I've noticed though is that with projects this large is that combining models makes it harder for a newcomer to hop in. It's one thing if we were able to combine everything in one or two reasonable files, but we would be at ~10ish still. With that many files, anyone not explicitly aware of the project structure will end up opening wrong things or resulting to a grep.

An alternative was to divide up the models into subfolders so that we could maintain the isolation  of resources. Seemed like that was just fragmenting further though, but I could be wrong.

YannickB:
Yep, I get the point for newcomers. Let's just do that.

PR #154, opened 2016-11-05
[IMP] clouder: Improved SSH handling

### LINKED ISSUE #152: [BUG] Stale SSH connection on remote server reboot
If the remote server is rebooted, the Clouder host server requires a restart to clear the stale connection in global memory. 

```
  File "/media/psf/LasLabs Inc/Repos/clouder/clouder/model.py", line 607, in execute
    channel = transport.open_session()
  File "/var/www/odoo/_venv/local/lib/python2.7/site-packages/paramiko/transport.py", line 705, in open_session
    timeout=timeout)
  File "/var/www/odoo/_venv/local/lib/python2.7/site-packages/paramiko/transport.py", line 789, in open_channel
    raise SSHException('SSH session not active')
SSHException: SSH session not active
```

Interesting thing is that the "Test Connection" button says that the connection is a success, so I think this might be broken

lasley:
This is a WIP version of the SSH environment I have been working on to fix #152. 

While debugging I also found an issue where multiple threads were able to utilize the same SSH connection at the same time, resulting in some incredibly interesting commands being sent to the remote node. This is also fixed in this PR by the use of thread locks upon command passthrough to the Paramiko client. 

The locks are currently only implemented at the connection level, but should be easy enough to add to the transport level as well. Doing this would allow multiple threads to use the same SSH connection, but different underlying transports. The caveat is a lot of downstream code would need refactoring in order to use transports instead of the raw connection, so I put this as a to do.

It seems work fine in console; left to do:

* Add tests
* Implement in models, refactor usage if clunky

Submitting this PR early in case there are comments regarding the implementation before I get deeper.

It should be noted that I made an `ssh` folder in the `clouder` module root. I plan on moving more SSH logic out of the model once the session handling has been completed.


lasley:
Note to self: it will currently yield an infinite loop of connections if server inaccessible

YannickB:
Hello @lasley, big work thank you. I also agree that ssh related code should be moved elsewhere because Clouder have different way to execute command, I also wanted to do it.

I agree with the design here, seems to go in the good direction.

lasley:
Alright this is working functionally. I will be adding tests, but won't be changing any of the code unless something comes up in a review or something is identified as broken in the tests. I'll squash up before we merge as well.

lasley:
Done and tested. I'm going to go over another round of deploys just to make sure we're good functionally but this is now good for review. I'll squash afterwards; I got lazy on my commits

lasley:
@t3ddftw please review

YannickB:
Looks good to me, I think I'll make more tests after merging but this seems to be going in the right direction.

And unit test.. thank you so much for this, this will be a life changer if we can trigger the oneclick in unit tests.

clouder/models/model.py
```diff
@@ -591,13 +625,13 @@ def execute(self, cmd, stdin_arg=False,
...
             cmd.append('"')
-            cmd.insert(0, '%s %s -c ' % (container, executor))
+            cmd.insert(0, '%s %s -c ' % (container, shell))
             if username:
                 cmd.insert(0, '-u ' + username)
             cmd.insert(0, 'docker exec')
 
         self.log('')
-        self.log('host : ' + host)
+        self.log('host : ' + ssh.host)
```
t3ddftw: `self.log('host : %s' % ssh.host)`

clouder/models/model.py
```diff
@@ -670,28 +704,31 @@ def get(self, source, destination, ssh=False):
...
 
-        host = self.name
-        if self._name == 'clouder.container':
-            # TODO
-            self.insert(0, 'docker exec ' + self.name)
-            host = self.server_id.name
+            host = record.name
+            if record._name == 'clouder.container':
+                # TODO
+                record.insert(0, 'docker exec ' + record.name)
```
t3ddftw: `record.insert(0, 'docker exec %s' % record.name)`

clouder/models/model.py
```diff
@@ -467,111 +477,135 @@ def connect(self, server_name='', port=False, username=False):
...
-            ssh = ssh_connections[host_fullname]
+        identity_file, host, username, port = self.__identity_file(
+            host, username, port
+        )
+
+        try:
+            env = SSHEnvironment(
+                host, port, username, identity_file,
+            )
+        except Exception as e:
```
t3ddftw: This seems a little broad. Shouldn't we just except for `SSHEnvironment.RETRY_EXCEPTIONS`?

YannickB:
On my side I am ready to merge it

lasley:
TBH I have no idea where or what @t3ddftw was commenting on because none of that code is in this PR.

Good to merge.

lasley:
Ohhhh wow there's a `Load Diff` button now - wtf I'm going to have to watch for that. 

We'll nail the interpolation in one swoop as previously determined in order to avoid a bunch of these changes randomly in out of scope PRs. Also the exceptions is on my list of things to do in a fell swoop too.

YannickB:
Ok let's merge then

PR #156, opened 2016-11-09
[IMP] Models for deployment refactoring

YannickB:
This PR add some models to manage the new concepts we want to manage, for review and comments before I start developing controllers : 

-Manage Docker Compose
-Manage Docker Swarm
-Deploy a node on Cloud (Amazon or other cloud providers) directly from Clouder
-Deploying containers directly on containers services (like Amazon ECS or Google container engine)
-Deploy in service now manage by container (like DNS, load balancing or backup)

It also add a first try for kubernetes runner. Unfortunately I didn't manage to make it works so I just commited my try and I'm moving to the next action (next one is docker compose)


YannickB:
In general I think it's good practice when we make important changes in core to first suggest models update through PR so we can review it before spending big time in developing the function behind.

clouder/models/config_settings.py
```diff
@@ -22,11 +22,21 @@ class ClouderConfigSettings(models.Model):
     email_sysadmin = fields.Char('Email SysAdmin')
     salt_master_id = fields.Many2one(
         'clouder.container', 'Salt Master', readonly=True)
+    deployer = fields.Selection(
+        [('engine', 'Docker Engine'), ('compose', 'Docker Compose'),
+         ('swarm', 'Docker Swarm'), ('runner', 'Custom Runner')],
```
lasley: IMO these select options should be generated via a method to allow injection without declaration. Or is the `runner` going to be for anything not explicitly covered?

clouder/models/provider.py
```diff
@@ -0,0 +1,29 @@
...
+
+class ClouderProvider(models.Model):
+
+    _name = 'clouder.provider'
+    _description = 'Provider'
+
+    name = fields.Char('Name')
+    config_id = fields.Many2one('clouder.config.settings',
+                                'Configuration', required=True)
+    type = fields.Selection(
```
lasley: `type` is a Python built in method, so we should try not to use it. I'm having a helluva time thinking of a better name though 😦

clouder/models/provider.py
```diff
@@ -0,0 +1,29 @@
...
+        [('instance', 'Instance'), ('container', 'Container'),
+         ('dns', 'DNS'), ('load', 'Load Balancing'), ('backup', 'Backup')],
+        string='Type', required=True)
+    access_id = fields.Char('Access ID')
```
lasley: Can we call this `access_id_external` or something? Ending in `_id` makes me think the col is going to be a *2o relation

clouder/models/provider.py
```diff
@@ -0,0 +1,29 @@
...
+    secret_key = fields.Char('Secret Key')
```
lasley: This string attribute is superfluous, default should already be this

clouder/models/provider.py
```diff
@@ -0,0 +1,29 @@
...
+class ClouderProvider(models.Model):
+
+    _name = 'clouder.provider'
+    _description = 'Provider'
+
+    name = fields.Char('Name')
+    config_id = fields.Many2one('clouder.config.settings',
+                                'Configuration', required=True)
+    type = fields.Selection(
+        [('instance', 'Instance'), ('container', 'Container'),
```
lasley: IMO these select options should be generated via a method to allow injection without declaration.

clouder/models/provider_template.py
```diff
@@ -0,0 +1,31 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class ClouderProviderTemplate(models.Model):
+
+    _name = 'clouder.provider.template'
+    _description = 'Provider Template'
+
+    @api.multi
+    def _get_images(self):
```
lasley: These methods should go below the field definitions

clouder/models/provider_template.py
```diff
@@ -0,0 +1,31 @@
...
+        return [('dummy', 'Dummy')]
+
+    @api.multi
+    def _get_sizes(self):
+        return [('dummy', 'Dummy')]
+
+    name = fields.Many2one('clouder.provider', 'Provider', required=True)
+    image = fields.Selection(_get_images, string='Image')
```
lasley: `lambda s: s._get_images()` to allow for inheritance without redeclaration

clouder/models/provider_template.py
```diff
@@ -0,0 +1,31 @@
...
+    size = fields.Selection(_get_sizes, string='Size')
```
lasley: `lambda s: s. _get_sizes()` to allow for inheritance without redeclaration

clouder_runner_kubernetes/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+
+{
+    'name': 'Clouder Runner Kubernetes',
+    'version': '9.0.10.0.0',
+    'category': 'Clouder',
+    'depends': ['clouder'],
+    'author': 'Yannick Buron (Clouder)',
+    'license': 'LGPL-3',
+    'website': 'https://github.com/clouder-community/clouder',
+    'demo': [],
```
lasley: Remove empty keys

clouder_runner_kubernetes/template.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
lasley: `<odoo>` and can remove data tag

clouder/models/config_settings.py
YannickB: Yes the runner is not explicit, the runner_id defined after it will allow you to specify for example an OpenShift or Kubernetes container which will take care of the deployment.

clouder/models/provider.py
YannickB: You have a point, but same here I have no good suggestion for now :/

clouder/models/config_settings.py
lasley: If we're using the runner_id for identification, why have the select at all then?

YannickB: Still I get the point, I'll do it

YannickB: Because Docker Engine / Compose / Swarm are integrated in the core module and will not be linked to a container at all. But all others runners shall be in a template module and linked to a container, so I think two fields are required.

PR #160, opened 2016-11-22
[IMP] Add support for Libcloud

YannickB:
This PR add support for Libcloud and allow to create an instance on the cloud (like Amazon) directly from the Clouder interface after filling credentials.

Tested with free amazon instance. 

Two bugs I was not able to resolve (it's Odoo related bug, not libcloud) : 
-The state of the instance is not displaying
-Same for size and location, they are not working correctly, that's why there is hard-coded values now.
We can fix theses bugs in another PR to avoid immediate functional review.


YannickB:
Ok I have trouble, I can't make travis go green because it can't see libcloud. Maybe because the name of the package (apache-libcloud) is not the same than import (libcloud). Any idea ?

YannickB:
Looks like I found a workaround. Ready for review.

lasley:
Woot yup the python-depends is what the import is. I swear that gets me so much, because it's totally backwards in comparison to setuptools or pip requires.

clouder/models/server.py
```diff
@@ -492,3 +562,94 @@ def clean_exec(self):
...
+    def libcloud_stop(self):
+        self = self.with_context(no_enqueue=True)
+        self.do('libcloud_stop', 'libcloud_stop_exec')
+
+    @api.multi
+    def libcloud_stop_exec(self):
+        cls = get_driver(getattr(Provider, self.provider_id.name))
+        driver = cls(self.provider_id.login, self.provider_id.secret_key)
+        for node in self.libcloud_get_nodes():
+            driver.ex_stop_node(node)
```
lasley: Generator ftw `driver.ex_stop_node(n for n in self.libcloud_get_nodes()`

.travis.yml
```diff
@@ -31,6 +32,7 @@ virtualenv:
   system_site_packages: true
 
 install:
+  - pip install apache-libcloud
```
YannickB: Damn I completely forgot this one xD

clouder/models/server.py
```diff
@@ -195,6 +211,60 @@ def _inverse_public_key(self):
...
+        locations = [('0', '0')]
+        if 'provider_id' in self.env.context and \
+                self.env.context['provider_id']:
+            provider = self.env['clouder.provider'].browse(
+                self.env.context['provider_id'])
+            cls = get_driver(getattr(Provider, provider.name))
+            driver = cls(provider.login, provider.secret_key)
+
+            locations = [(l.id, l.id) for l in driver.list_locations()]
+            _logger.info('%s', locations)
```
YannickB: this logger is only here until we understand why his value does not display in the selection field. Will be removed after it.

YannickB:
Ready for review

lasley:
@t3ddftw please review

clouder/models/provider.py
```diff
@@ -8,26 +8,36 @@
 from openerp import fields
 from openerp import models
 
-
 _logger = logging.getLogger(__name__)
 
+try:
+    import libcloud
+except ImportError:
+    _logger.warning('Cannot `import libcloud`.')
```
t3ddftw: Hmm, wouldn't we want to up the log level here to error? Given that the existence of this library is crucial to this module functioning.

clouder/models/server.py
```diff
@@ -2,17 +2,24 @@
...
-
-import logging
 _logger = logging.getLogger(__name__)
 
+try:
+    from libcloud.compute.base import NodeAuthSSHKey
+    from libcloud.compute.providers import get_driver
+    from libcloud.compute.types import Provider
+except ImportError:
+    _logger.warning('Cannot `import libcloud`.')
```
t3ddftw: Again, I believe we should up the log level to error.

clouder/models/provider.py
lasley: No, convention is warning. Odoo module loader will hit all imports at the top of files, whether or not it is installed. Therefore, this may not be an error.

clouder/models/server.py
lasley: 👎

YannickB:
Ok then, just merging recent commit from master and if everyone is okay let's merge it.

YannickB:
All checks ok and I suppose everybody agree, I'm merging

PR #158, opened 2016-11-11
[IMP] Add support for Docker Compose and Docker Swarm

YannickB:
First work for Docker Compose and Docker Swarm. I managed to deploy with it the container Salt Master, but I still have to make a debug on the whole oneclick.

The support for Docker Compose is only experimental because it currently not work yet with Docker Swarm. Still I made a first work for it because of all the challenges it implies to manage this concept (Currently Engine and Swarm deploy containers one by one at the deepest level, while a Compose deployment require a deployment at the upper level) and I'm quite proud of it, should not worry too much for the future now.
Also Salt is deactivated for now until we need it, want to focus on Swarm debug and Salt has no plugin for Swarm yet.

The default Clouder setup will become Swarm/No Compose/SSH. In the end (when Docker and Salt will allow it), the default setup shall become Swarm/Compose/Salt but for now the other setup shall do it.

PR #163, opened 2016-11-22
[FIX] clouder: Remove dup view field

lasley:
This fixes a bug introduced in #158 that I didn't notice during review because I didn't set the server up fresh on functional test (my bad!). Duplicating the field generates a constrain issue when saving:

```
Error while validating constraint

expected string or buffer
```

* Remove duplicated public IP from clouder server view


YannickB:
Can you check if this PR is still needed ?

lasley:
This is solved on master now

PR #164, opened 2016-11-23
[IMP] Alpine

YannickB:
And finally the last PR

I made an almost complete oneclick deploy, should still be a little buggy though. Proxy and Odoo/postgres functional, not bind nor postfix.

Please do not review the Dockerfiles too much, especially the considerable amount of commented code. I want to keep the old RUN until we are sure that all images works perfectly under alpine.


clouder/images/backup-bup/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM yannickburon/clouder:base
```
lasley: One note now that I'm looking at these Dockers, even though I said I wasn't going to. We should be versioning our images inline with the version of software they repesent. Eg this would be `clouder-base:3.4` assuming Alpine 3.4, and Nginx would be `clouder-nginx:1.10.2` assuming current stable

YannickB:
Ok so let's merge it, I'll start configuring Docker Hub with the new Dockerfile location and then make some tests.

PR #165, opened 2016-11-27
[IMP] clouder_template_proxy: Improve nginx configs

lasley:
We should probably wait for #164 merge & test this again.

* Centralize security and proxy settings
* Upgrade SSL settings to current Mozilla recommended for intermediate compatibility
* Add ipv6 support
* Add default_server to non-root proxy config

Looks like we'll also need to start updating the other proxy configs, but I figured this could come first.


clouder/models/base.py
```diff
@@ -897,3 +899,22 @@ def renew_cert_exec(self):
         Renew a certificate
         """
         return True
+
+    @api.multi
```
YannickB: I think this function shall be in the proxy template, especially since it execute system command in the proxy container

YannickB:
Ok, ready for merge ?

lasley:
Yup good to go!

PR #168, opened 2016-11-29
[IMP] Replace with Docker Hub image and Debug oneclick

YannickB:
Still to do : 
- Oneclick end with connexion error with deploying Odoo base
- DNS to refactor
- Postfix not working


YannickB:
Ok I did something really stupid, this PR is still open please review (cc @lasley)

clouder_template_postgres/template.xml
```diff
@@ -33,7 +33,7 @@
         <record id="image_postgres_data" model="clouder.image">
             <field name="name">image_postgres_data</field>
             <field name="template_ids" eval="[(4, [ref('image_template_postgres_data')])]"/>
-            <field name="parent_from">yannickburon/clouder:postgres-data</field>
+            <field name="parent_from">clouder/postgres-data:0.10.1</field>
```
lasley: We would want these tags to be the version of software that the image is representing. In the case of Postgres likely `9.6`.

clouder/models/container.py
```diff
@@ -1288,7 +1288,8 @@ def recursive_deploy_one(self):
         """
 
         if self.child_ids:
-            self.child_ids.mapped('child_id').recursive_deploy_one()
+            for child in self.child_ids:
+                child.mapped('child_id').recursive_deploy_one()
```
lasley: The `mapped` should keep us from having to loop. Were you running into a singleton error or something? If so, `child.child_id` would be functionally equivalent without having to call `.mapped`

YannickB: yeah I experimented singleton error

I'm not sure of what gonna happen if child.child_id is False, but ok let's try it

clouder_template_postgres/template.xml
YannickB: I am not completely sure about that, because we may do some change in Dockerfile between Clouder version, and in this case we don't want two clouder version to use the same image version if the software version didn't changed. I think clouder version is still the way to go.

YannickB:
Fast merging this PR, I need it to correctly test the container -> service PR

Issue #159, opened 2016-11-17
[REF] Rename some core concepts


YannickB:
I think we need to rename some of Clouder core concept, to better fit with name used in other Docker related tools : 

server -> node
container -> service
save -> backup

Of course theses name are used everywhere in Clouder so this is not so trivial. I'll make one PR for each change.

Any comment before I start working on it ?

YannickB:
All done I think, I close

PR #169, opened 2016-11-29
Rename container into service

YannickB:
#159


YannickB:
Yep this kind of rename usually generate unexpected and funny errors, but that's not the first time I do this kind of things and usually it goes quite well. Just did search and replace, it's the most reliable in such case.

Still, I expect the rename in some comments will be funny, where service will just does not make any sense in sentence.

I am merging it without test, I wanted to do somes but the previous PRs introduce errors and clearly we should not correct them in this massive one-goal PR

PR #171, opened 2016-11-30
[REF] Rename server into node

YannickB:
#159


YannickB:
Yeah this one was a little more trickier, server was especially used in many sources and Dockerfiles. I made the search and replace only on the *.xml and *.py so I think I avoided most of the wrong replace, and I reviewed each change manually.

YannickB:
It shall be noted that I made a oneclick installation with the new service and node terms, so no need to worry too much we shall just have maybe some random errors here and there in the next weeks.

Issue #157, opened 2016-11-11
[RFC] Billing and Deploy Flow


lasley:
# Overview

I've been thinking about this for a while, and I think I have a plan for billing Clouder instances. We have a lot of framework built already, but it's duplicating a lot of what Odoo can already do for us. My goal will be to offload as much logic out of our scope as possible.

For this plan to work, we need a job queue. I will assume that we will depend on either `connector` or the base `queue_job` in order to accomplish this, but I guess it could also be built from scratch to allow the LGPL licensing on this module. Lots of work though.

From what I'm seeing/thinking, we have three independent concepts for billing:
* Threshold based - Customer pays for X & receives no more than X without explicit upgrade
* Usage based - Customer pays for X per Y units
* Flat fee - Customer pays X, with no metering

Whatever we decide to actually bill on will likely be a combination of all three types of billing concepts. Some of these fees would repeat, and some would be one-time. Both pre- and post-billing cycles are represented in these concepts.

We also need some of these fees to come with an action, such as:
* Instance creation 
* Instance destruction
* Sending appropriate commands for threshold limits

# Sale

Each of the fees will be a `product.product`, just like normal Odoo. We will need to add the following onto `product.product`:
* Invoicing Policy Types - `Recurring`, `Threshold`, `Usage`
* `clouder_application_ids` - M2M with `clouder.application`. Used for manufacturing
* `recur_interval` - In months

Products representing an instance itself will be set to go through an `Infrastructure Manufacture` Warehouse Route, and have an associated Bill of Materials. We can even dynamically set the manufacturing lead time based on average (or upper quartile) build time for that instance type.

# Manufacture

Each server, or maybe runner, will be a Manufacturing Work Center. We could figure out a way to determine the efficiency factor & capacity on the fly, but that's not important now. I think that we would need to add `Infrastructure` for use as the Resource Type of the Work Center.

When a Product is purchased, the Bill of Materials will generate the Stock Picking with the products that are contained as part of the kit. Creation of a Stock Picking in the `Infrastructure Manufacture` Route will trigger a Job Queue addition for all Products in the Picking. The sequence of the Picking will be respected, which will determine the build order. This sequence is the same as in the Bill of Materials for the Product.

# Bill

## Recurrence

Use new `recur_interval` on Product to create a `subscription.document`, or identify an existing `subscription.document` on the same interval to append to. This would be triggered upon either confirmation of the sale, or perhaps the validation of the first invoice.

[...]

PR #173, opened 2016-12-01
[WIP] Create metric and billing interface outline

lasley:
This is a prelim outline of what I have been describing in #157. Nothing functional at the moment other than data schemas and a few theoretical methods.

There's also a shell of a metric interface definition, but that was just so I had something to connect the billing models to. I'm still coming up with a real plan there.


sale_clouder/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+from openerp import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    invoice_policy = fields.Selection(
+        selection_add=[
+            ('threshold', 'Invoice and Enforce a Threshold'),
+            ('usage', 'Invoice Based on Usage'),
```
lasley: This may be removed in favor of using pre-existing `delivery` policy, which represents the same thing for the most part

YannickB:
Great @lasley , I had a first look at it but I'll have a second one tomorrow. Thanks!

sale_clouder/models/clouder_contract_line.py
```diff
@@ -0,0 +1,31 @@
...
+    _inherits = {'account.analytic.invoice.line': 'contract_line_id'}
+
+    contract_line_id = fields.Many2one(
+        string='Recurring Line',
+        comodel_name='account.analytic.invoice.line',
+        index=True,
+        required=True,
+        ondelete='restrict',
+    )
+    service_id = fields.Many2one(
```
YannickB: Don't forget that we shall also be able to invoice a base, not only a service.

clouder_metric/models/clouder_metric_value.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ClouderMetricValue(models.Model):
```
YannickB: I don't really understand the use of this model, because we said the value would be recovered realtime from ELK. If we need to store it here, we shall probably have a datetime field to know when the value was recovered.

That said, I understand this is a POC and it will probably change when we get close to the end.

lasley: Yeah everything is going to be stored in ELK, but we need some way to store the aggregate billing metrics so they don't change out from under us after they have been used for billing.

Yup you're totally right about the need for datetimes indicating the aggregation interval. I added that in my RFC at https://github.com/clouder-community/clouder/issues/174

This module is likely going to be removed from this PR, it was just here so I could connect accounting to something while figuring that all out.

sale_clouder/models/clouder_contract_line.py
lasley: Ah good call, I knew it seemed like I was missing a layer of abstraction!

lasley: Just so I'm clear, what metrics are going to be monitored on base? That's just going to be flat rate for the most part right?

lasley: Hmmmm but wouldn't the number of users be dictated in the Service itself? The base represents a collection of services doesn't it?

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,145 @@
...
+        Returns:
+            (int) Quantity to use on invoice line in the UOM defined on the
+                ``contract_line``.
+        """
+        raise NotImplementedError
+        self.get_quantity_flat(account_line, invoice)
+
+    @api.multi
+    def _get_quantity_usage(self, account_line, invoice):
+        """ It provides a quantity based on unbilled and used metrics """
```
tedsalmon: Hey @lasley,

My assumption is that we need to get a RecordSet of `clouder.metric.interface` to return here, right? I'm just failing to find a good relation between the two. Perhaps it just wasn't implemented yet?

lasley: Ummm no, it would be an actual numeric. All of these `_get_quantity_*` methods would be used to determine the `product_qty` of the invoice line item.

lasley: Take a look at _get_quantity_flat

tedsalmon:
Hey guys,

This is should be ready for review. Let me know if I missed anything.

Thanks!
-Ted

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @api.model
+    def _create_default_vals(self, account):
+        """ It returns default values to create and link new ClouderContracts.
+
+        Args:
+            account (account.analytic.account): Account that ClouderContract
+                will reference.
+        Returns:
+            (dict) Values fed to ``create`` in ``_get_contract_by_account``.
```
lasley: `dict: Values`

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        number = self.env['ir.sequence'].next_by_code('clouder.contract')
+        # Default to the current users company if not set
+        company_id = account.company_id.id or self.env.user.company_id.id
```
lasley: This will fail if no `account.company_id`

```
account.company_id and account.company_id.id or self.env.user.company_id.id
```

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+    @api.model
+    def _get_contract_by_account(self, account, create=False):
+        """ It returns the ClouderContract or possibly creates a new one.
+
+        Args:
+            account: (account.analytic.account) Contract to search by.
+            create: (bool) True will create a new ClouderContract if one does
+                not already exist.
+        Returns:
+            (clouder.contract) Clouder contract associated with ``account``.
```
lasley: `ClouderContract: Clouder`

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+        adding the following into the ``code`` field:
+
+        .. code-block:: python
+
+        result = env['clouder.contract'].get_invoice_line_quantity(
+            contract, line, invoice,
+        )
+
+        Args:
+            account (account.analytic.account): Contract that recurring
```
lasley: These model names should be switched for class names instead. e.g. `account.analytic.account` == `AccountAnalyticAccount`

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @api.multi
+    def _get_quantity_flat(self, account_line, invoice):
+        """ It returns the base quantity with no calculations
+        Args:
+            account_line (account.analytic.invoice.line): Recurring invoice
+                line being referenced.
+            invoice (account.invoice): Invoice that is being created.
+        Returns:
+            (float) Quantity with no calculations performed
```
lasley: `float:`

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+    @api.multi
+    def _get_quantity_threshold(self, account_line, invoice):
+        """ It functions like flat rate for the most part
+
+        Args:
+            account_line (account.analytic.invoice.line): Recurring invoice
+                line being referenced.
+            invoice (account.invoice): Invoice that is being created.
+        Returns:
+            (float) Quantity to use on invoice line in the UOM defined on the
```
lasley: `float:`

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+        start_date = inv_date - inv_delta
+
+        # Filter out metrics that fall within the billing period
+        def filter(rec):
+            if not all((rec.date_start, rec.date_end)):
+                return False
+            start = fields.Datetime.from_string(rec.date_start)
+            end = fields.Datetime.from_string(rec.date_end)
+            return start >= start_date and end <= inv_date
```
lasley: Add blank line

sale_clouder/models/clouder_contract.py
```diff
@@ -0,0 +1,180 @@
...
+        usage_values = vals.filtered(filter)
+        for val in usage_values:
```
lasley: `usage = sum(usage_values.mapped('value'))`

sale_clouder/models/clouder_contract_line.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ClouderContractLine(models.Model):
+    """ It provides the link between billing and Clouder Services. """
+
+    _name = 'clouder.contract.line'
```
lasley: _description

tedsalmon:
Thanks @lasley! I've attended to your comments in commit 8c7645c & 400f151 

Again, I was unable to switch the `fields.Selection` field to `fields.Reference` because the latter is a reference to a specific record of a model; not the model itself.

I also boiled down some of the logic used to calculate the usage quantity in `clouder.contract`

Thanks!

clouder_metric/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
YannickB: Argh... This may be a problem, the metric module will probably be deployed in all infrastructures. If it becomes AGPL, we can practically assume that all Clouder will become AGPL, this is not the same case than clouder_asynchronous.

I'm afraid I have to be against such move just because we use this dependency. Either we can convince people that contract_variable_quantity is considered a component and shall be LGPL, or we'll have to do without it.

YannickB: That being said, I think I'll not block the PR just for that, we can find a solution later and merge it AGPL for the time being.

YannickB:
Hi guys,

Good job ! I'll make my review this afternoon, hope we can merge it quickly after that. Thanks again for your hard work.

sale_clouder/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
YannickB: Please make a symbolic link from __openerp__.py to __manifest__.py to ensure the module can be installed on Odoo 9 (see my last commit on master)

clouder_metric/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
YannickB: Please make a symbolic link from __openerp__.py to __manifest__.py to ensure the module can be installed on Odoo 9 (see my last commit on master)

clouder_metric/__manifest__.py
lasley: @YannickB - I think we'll be fine here. Reason being is that the metrics and sales modules are far from the core, nothing will need to depend on them, and there will likely not be third party proprietary changes. 

Anything to do with contract is licensed AGPL, primarily due to it being ripped out from v8 - before the LGPL license switch in v9. I don't really want to rewrite the contract module at this time, but that would be the requirement to move from this.

clouder_metric/__manifest__.py
lasley: I'm not so sure about this. What about the changed namespace (`from odoo` vs `from openerp`)? I'm pretty sure I have also used some new API decorators that will not work on v9 (`api.model_cr_context`)

clouder_metric/__manifest__.py
YannickB: Very well, I'm fine anyway we'll figure out what happen later.

clouder_metric/__manifest__.py
YannickB: Well at the very least the tricks worked on my dev instance, and all other modules now have the symbolic link. Anyway I'll not block the PR for this, so I'm just waiting for your approval.

lasley:
LGTM thanks @tedsalmon. We'll likely need to make some changes when we actually implement with the Elastic connector, but this is a great start for a merge IMO.

YannickB:
OK, LGTM for me too then. I'm merging, thanks @tedsalmon @lasley for the hard work !

PR #175, opened 2016-12-01
[REF] DNS

YannickB:
- Change module into clouder_template_dns

- Add support for libcloud DNS (untested yet)

Annnnd I'd like to say debug bind, but no it's not working yet... It's starting to be a problem and a priority I need help, I don't understand were the problem is coming (even an nslookup from inside the bind container is not working). The cause seems to be the move to alpine.
I need a bind guru to look at it, and if possible review the entire configuration because I'm pretty sure it's insecure, I am really bad with DNS.

Still, I suggest to merge this PR for the design improvement and debug bind on another one


YannickB:
@lasley I'll need your help here I think, I need to merge this PR and another one before I can continue

lasley:
Ooops I missed this, will review today

clouder_template_dns/__init__.py
```diff
@@ -21,5 +21,7 @@
 #
 ##############################################################################
 """
+
+from . import common
 from . import bind
-from . import route53
+from . import clouddns
```
YannickB: I agree, but I had some concern with the application_type having an underscore. Thus, this require further tests.

I don't want to block the debug branch for this, so I'll merge anyway and correct it later.

PR #177, opened 2016-12-07
[REF] Refactor mail container

YannickB:
Rename into clouder_template_mail

Now use external provider object for smtp relay (even if we don't use libcloud here)

Reactivate opendkim (testing package in alpine)

Still not completely tested, but let's say it's ready for debug


YannickB:
Damn, I commit and left without checking Travis, I was sure it'd work :(

lasley:
Travis is 🍏 , so the cross-merge seems to have done something good 👍

PR #178, opened 2016-12-08
[IMP] Change localport to local_port

lasley:
I noticed this while doing the Elastic template & it's killing my OCD. Cannot unsee!

Change should be safe enough, this was just a simple `sed`. That's assuming that the no snake casing was not for namespace conflict avoidance similar to `clouddns`


YannickB:
Ok let's merge. Note there is probably the same with hostport.

PR #179, opened 2016-12-14
[ADD] clouder_template_elasticsearch: Provide Elastic

lasley:
This provides a base Elasticsearch 5.0 template. 

The config file is pretty blank, but we can hit that in another ticket after I finish tuning it on my dev install.

Regarding the backups - we have to use another tool called [Elasticdump](https://www.npmjs.com/package/elasticdump). It's best to run that in a container of its own though, so I just added to the readme.

Did I miss anything else? I feel like I missed something.


clouder_template_elasticsearch/data/application.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 LasLabs Inc.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
```
lasley: Used wrong license, switch to LGPL

clouder_template_elasticsearch/images/5.0-data/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM yannickburon/clouder:base
```
YannickB: Are you sure we shall have the versioning number (5.0) in the path ? According to me we with clouder branches and configuration in docker hub we already have what we need.

Moreover, I'd like to keep the same name in path image than the name which will appears on Docker hub, thus clouder_template_elasticsearch/images/elasticsearch-data/Dockerfile

YannickB: FROM clouder/base:3.4

clouder_template_elasticsearch/images/5.0-exec/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM yannickburon/clouder:base
```
YannickB: FROM clouder/base:3.4

clouder_template_elasticsearch/models/backup.py
```diff
@@ -0,0 +1,28 @@
...
+
+    All methods and properties are to be prefixed with ``elastic_`` in order
+    to prevent namespace clashes with existing operations, unless overloading
+    and calling + returning the super.
+    """
+
+    _inherit = 'clouder.backup'
+
+    @api.multi
+    def backup_database(self):
```
YannickB: We'll probably call deploy_base function like in https://github.com/clouder-community/clouder/blob/0.9.0/clouder_template_odoo/template.py#L561 and not backup_database.

Elasticdump will need to put files in /base-backup like we do for odoo filestore. Except if the basic container save already do the trick, maybe we don't need elacticdump.

clouder_template_elasticsearch/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Clouder Template Elasticsearch',
+    'version': '9.0.10.0.0',
+    'category': 'Clouder',
+    'depends': [
+        'clouder',
+        'clouder_template_proxy',
```
YannickB: Since it's a web application, some dependences are probably missing, like dns

clouder_template_elasticsearch/data/application_type.xml
```diff
@@ -0,0 +1,17 @@
...
+
+<odoo>
+
+    <record id="application_type_elasticsearch"
+            model="clouder.application.type"
+            >
+        <field name="name">elasticsearch</field>
+        <field name="system_user">elasticsearch</field>
+        <field name="tag_ids"
+               eval="[(6, 0, [ref('clouder.tag_database')])]"
```
lasley: Hmmm but doesn't Elasticsearch meet the definition of a database? I admittedly have no idea what this is actually doing, but it seemed a good idea to make this a database interface. 

From there we would attach the Logstash instance (for example) via similar mechanisms used for attaching databases to, say, an Odoo instance. Am I overthinking it?

clouder_template_elasticsearch/data/image_port.xml
```diff
@@ -0,0 +1,23 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+
+<odoo>
+    
+    <record id="image_port_elasticsearch_http"
+            model="clouder.image.port"
+            >
+        <field name="template_id" ref="image_template_elasticsearch_exec" />
+        <field name="name">elasticsearch-http</field>
+        <field name="local_port">9200</field>
```
lasley: Is my assumption that switching to `https` would identify it as an SSL only port?

clouder_template_elasticsearch/images/5.0-data/Dockerfile
lasley: I think the explicit versioning will be good here - Elastic has historically not been very loving to old API between major versions, so the difference between them is fairly drastic. 

An upgrade of Elastic sometimes requires a full cluster restart or a reindex, and sometimes things simply don't work after the fact. A great example is the current fracturing of Kibana dashboards - almost nothing works out of the box, because they're for 2.x.

Due to this, I think we need to manage Elastic major versions in the same respect we would an Odoo version. Should this instead be `clouder_template_elasticsearch_5`, then drop versioning on inner folder hierarchy?

clouder_template_elasticsearch/models/backup.py
lasley: The container save should work now that you say it. I have both the config and the db data stored on the data container, so simply adding another exec onto a copy of it should yield an identical installation assuming that the exec versions are the same.

clouder_template_elasticsearch/data/application_type.xml
YannickB: Maybe the definition is the same, but we're not gonna use it the same way than postgresql or mysql (like, the place where the database of clouder.base is stored).

The database tag trigger some specific code like https://github.com/clouder-community/clouder/blob/master/clouder/models/service.py#L135, so it's best not to mix them. We'll see later if the database tag is really needed.

clouder_template_elasticsearch/data/image_port.xml
YannickB: yep exactly https://github.com/clouder-community/clouder/blob/master/clouder_template_proxy/template.py#L253

clouder_template_elasticsearch/images/5.0-data/Dockerfile
YannickB: Ok I see the point and the comparison with Odoo template. Then, I suggest the path 
clouder_template_elasticsearch/images/elasticsearch-data/Dockerfile clouder_template_elasticsearch/images/elasticsearch5-exec/Dockerfile 

Exec need to be versionned but not data IMO (same for Odoo template, data isn't versionned)

lasley: Hmm but in this case, the data itself could become invalidated by the upgrade. One main thing I've been running into with the Kibana Dashboards has been the removal of support for a specific data type.

I guess this is similar to the Odoo data though, which would become invalidated with a new exec.

I'm curious then - what is the upgrade path for the non-versioned data? Some of this also looks to contain configs, which could change by a version.

YannickB: That's what I was thinking also, we still need to think about how to manage major migration in Clouder. For example I'd love to have automatic Odoo migration using OpenUpgrade (and I guess I'm not the only one..).

I suggest to stick with current version for now, and get back to it when we would be able to manage major Odoo migration in Clouder.

lasley: Yeah that's a definite must. Ok yeah so I'm going to version both the data and exec on this one in order to avoid future changes then. Will use your recommended naming of `elasticsearch5-*`

lasley:
@YannickB Alright this is good for a second review and possible merge.

I have another ticket open to add SSL between Elastic and the Proxy.

Did I need to add any proxy configs, or are those all automatic assuming the use of the `http` tag on the port?

YannickB:
OK LGTM.

The proxy will automatically use the http tag, so no configuration needed here.

PR #180, opened 2016-12-14
[ADD] clouder_certificate_authority: Implement CA using CFSSL

lasley:
This is a WIP - I still need to implement an interface to the CLI. I'm having a real helluva a time getting this to work the way I want though, so maybe you can help.

CFSSL isn't a service like everything else is, it's a CLI app that is hosted in its own Docker container. See the following console:
```
$ docker run -i -t lasley/cfssl-exec
Usage:
Available commands:
	serve
	ocspsign
	ocspserve
	selfsign
	bundle
	genkey
	gencert
	certinfo
	scan
	info
	print-defaults
	revoke
	version
	gencrl
	ocspdump
	ocsprefresh
	sign
Top-level flags:
  -allow_verification_with_non_compliant_keys
    	Allow a SignatureVerifier to use keys which are technically non-compliant with RFC6962.
```

I tried setting `clouder.application.link.service` to false, but that didn't seem to have any affect. Is what I'm looking to do supported?


clouder_certificate_authority/images/1.2.0-exec/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+	&& rm -rf /go \
+	&& echo "Build complete."
+
+# Create and Change to PKI Dir
+RUN mkdir -p /var/pki
+WORKDIR /var/pki
+
+# Enter
+ENTRYPOINT ["cfssl"]
+CMD ["--help"]
```
YannickB: This means the command of your Dockerfile is `cfssl --help`

This command is not permanent, it immediately exit, thus killing your container.

If you just want to execute command inside this container and not provide any services with it, while keeping it running, just don't define any ENTRYPOINT nor CMD.
Like any data container in Clouder, it'll use the CMD defined in the base container, aka `tail -f /dev/null` which keep the container running.

clouder_certificate_authority/images/1.2.0-exec/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM yannickburon/clouder:base
```
YannickB: I forgot to make the change on most Clouder images, but you can use the official repo now

`FROM clouder/base:3.4`

clouder_certificate_authority/data/application_link.xml
```diff
@@ -0,0 +1,20 @@
...
+            >
+        <field name="template_id"
+               ref="application_template_cfssl"
+               />
+        <field name="name" ref="application_cfssl" />
+        <field name="service" eval="False" />
+        <field name="required" eval="True" />
+        <field name="auto" eval="True" />
+        <field name="make_link" eval="True" />
+    </record>
```
YannickB: I don't think you need this link.

The link feature is used when you want to execute action depending on other component of the infrastructure.
Example : a link between Odoo container to postgres container so odoo user are created on postgres container, the user is deleted and recreated each time we redeploy this link.

In your case, I believe for example if you need a certificate for the elactic container, you'll want a link in elastic container to the cfssl, which will generate the certificate for elastic on deploy. But I don't see the point of a link cfssl to cfssl :)

clouder_certificate_authority/data/application_type_option.xml
```diff
@@ -0,0 +1,29 @@
...
+    <record id="application_type_option_cfssl_listen"
+            model="clouder.application.type.option"
+            >
+        <field name="application_type_id"
+               ref="application_type_cfssl"
+               />
+        <field name="name">listen</field>
+        <field name="type">service</field>
+        <field name="default">*</field>
+    </record>
```
YannickB: I feel like you copied the options from the postgres container, I don't believe you'll need theses options

clouder_certificate_authority/models/certificate_name.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import api, fields, models
+
+
+class ClouderCertificateName(models.Model):
```
YannickB: Creating a new model, I'll keep an eye on what you're doing to make sure this model is really needed.

I think it may be, and thus will replace the cert_key and cert_public fields in clouder.base, clouder.domain by many2one fields.

Waiting to see views to be sure.

clouder_certificate_authority/images/1.2.0-exec/Dockerfile
lasley: Got it, so the tail keeps the container available so that I can then call it with `clouder.model.execute` with the args that I would want to send to my shell script?

clouder_certificate_authority/data/application_type_option.xml
lasley: I totally did 😆

clouder_certificate_authority/models/certificate_name.py
lasley: Yeah so this is my plan, and I'm actually planning some sort of RFC or email or something. We've got a problem here that I'm not sure can be solved in Odoo.

clouder_certificate_authority/images/1.2.0-exec/Dockerfile
YannickB: yep exactly

clouder_certificate_authority/models/certificate_name.py
YannickB: ok, waiting for more information

clouder_certificate_authority/images/1.2.0-exec/Dockerfile
lasley: Sweet this makes sense, thanks.
Side note- I wonder what the performance impact of leaving the tail running is vs using it as a console app is. Have you measured this at all?

lasley: @t3ddftw - thoughts? We're obviously counting pennies, but good to think about given the goals of this project are centered around scaling.

t3ddftw: What about running `cat` without an argument? That will have the same result but without taking up a file handle AND it is seven less syscall's than `tail -f /dev/null`.

`tail`:
```
execve("/bin/tailf", ["tailf", "/dev/null"], [/* 17 vars */]) = 0
brk(0)                                  = 0x151b000
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f30715dc000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=24043, ...}) = 0
mmap(NULL, 24043, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f30715d6000
close(3)                                = 0
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
open("/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0P \2\0\0\0\0\0"..., 832) = 832
fstat(3, {st_mode=S_IFREG|0755, st_size=1840928, ...}) = 0
mmap(NULL, 3949248, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f3070ff7000
mprotect(0x7f30711b1000, 2097152, PROT_NONE) = 0
mmap(0x7f30713b1000, 24576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x1ba000) = 0x7f30713b1000
mmap(0x7f30713b7000, 17088, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7f30713b7000
close(3)                                = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f30715d5000
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f30715d3000
arch_prctl(ARCH_SET_FS, 0x7f30715d3740) = 0
mprotect(0x7f30713b1000, 16384, PROT_READ) = 0
mprotect(0x603000, 4096, PROT_READ)     = 0
mprotect(0x7f30715de000, 4096, PROT_READ) = 0
munmap(0x7f30715d6000, 24043)           = 0
brk(0)                                  = 0x151b000
brk(0x153c000)                          = 0x153c000
open("/usr/lib/locale/locale-archive", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=2919792, ...}) = 0
mmap(NULL, 2919792, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f3070d2e000
close(3)                                = 0
stat("/dev/null", {st_mode=S_IFCHR|0666, st_rdev=makedev(1, 3), ...}) = 0
open("/dev/null", O_RDONLY)             = 3
fstat(3, {st_mode=S_IFCHR|0666, st_rdev=makedev(1, 3), ...}) = 0
ioctl(3, SNDCTL_TMR_TIMEBASE or SNDRV_TIMER_IOCTL_NEXT_DEVICE or TCGETS, 0x7fff83f8afb0) = -1 ENOTTY (Inappropriate ioctl for device)
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f30715db000
read(3, "", 4096)                       = 0
close(3)                                = 0
munmap(0x7f30715db000, 4096)            = 0
inotify_init()                          = 3
inotify_add_watch(3, "/dev/null", IN_MODIFY|IN_DELETE_SELF|IN_MOVE_SELF|IN_UNMOUNT) = 1
read(3,
```

`cat`:
```
execve("/bin/cat", ["cat"], [/* 17 vars */]) = 0
brk(0)                                  = 0x1c31000
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f29ec71d000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=24043, ...}) = 0
mmap(NULL, 24043, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f29ec717000
close(3)                                = 0
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
open("/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0P \2\0\0\0\0\0"..., 832) = 832
fstat(3, {st_mode=S_IFREG|0755, st_size=1840928, ...}) = 0
mmap(NULL, 3949248, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f29ec138000
mprotect(0x7f29ec2f2000, 2097152, PROT_NONE) = 0
mmap(0x7f29ec4f2000, 24576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x1ba000) = 0x7f29ec4f2000
mmap(0x7f29ec4f8000, 17088, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7f29ec4f8000
close(3)                                = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f29ec716000
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f29ec714000
arch_prctl(ARCH_SET_FS, 0x7f29ec714740) = 0
mprotect(0x7f29ec4f2000, 16384, PROT_READ) = 0
mprotect(0x60a000, 4096, PROT_READ)     = 0
mprotect(0x7f29ec71f000, 4096, PROT_READ) = 0
munmap(0x7f29ec717000, 24043)           = 0
brk(0)                                  = 0x1c31000
brk(0x1c52000)                          = 0x1c52000
open("/usr/lib/locale/locale-archive", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=2919792, ...}) = 0
mmap(NULL, 2919792, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f29ebe6f000
close(3)                                = 0
fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 1), ...}) = 0
fstat(0, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 1), ...}) = 0
fadvise64(0, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
read(0, 
```

`yes` endlessly writes to `stdout`, so I can see how that ended up being resource intensive :)

YannickB: Well why not, cat looks good

clouder_certificate_authority/data/application_link.xml
YannickB: Yep, exactly

clouder_certificate_authority/cfssl.py
```diff
@@ -0,0 +1,381 @@
+# -*- coding: utf-8 -*-
```
lasley: I'm moving this to https://github.com/laslabs/cfssl-py

clouder_certificate_authority/data/application.xml
```diff
@@ -0,0 +1,56 @@
...
+        <field name="name">CFSSL Exec</field>
+        <field name="code">exec</field>
+        <field name="type_id" ref="application_type_cfssl" />
+        <field name="tag_ids" eval="[(4, ref('clouder.tag_exec'))]" />
+        <field name="default_image_id" ref="image_cfssl_exec" />
+        <field name="sequence">2</field>
+        <field name="required" eval="True" />
+        <field name="update_strategy">auto</field>
+        <field name="child_ids"
+               eval="[(4, ref('application_openssl_exec'))]"
```
lasley: If I nest an exec in an exec like this, will the deploy recurse correctly?

clouder_certificate_authority/api.py
```diff
@@ -0,0 +1,41 @@
...
+
+        Args:
+            certificate_authority (:type:`clouder.CertificateAuthority`):
+                The certificate authority record singleton representing the
+                remote API. The CA does not have to be initialized yet. Use
+                :type:`None` if ``self`` is the CA that should be connected
+                to.
+        """
+        try:
+            # @TODO: Figure out how the hell to get this host from the base
```
lasley: I need the IP of the container and exposed port here in order to access the web API. Or maybe I need the IP/port of the proxy instead?

clouder_certificate_authority/data/application.xml
YannickB: Hum, you're creating a tree like this : 
cfssl
-cfssl data
-cfssl exec
--openssl exec

I suspect we'd want to have this tree : 
cfssl
-openssl exec
-cfssl data
-cfssl exec
And make a link between openssl exec and cfssl exec

More information incoming when I'll make a proper review of the PR (this day I hope)

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+
+from ..api import API
+
+
+class ClouderCertificateAuthority(models.Model, API):
+    """ It provides an interface for controlling a Cert Authority. """
+
+    _name = 'clouder.certificate.authority'
+    _description = 'Clouder Certificate Authority'
+    _inherits = {'clouder.application': 'application_id'}
```
YannickB: I believe the main error is here : According to me ClouderCertificateAuthority shall inherit clouder.service, not clouder.application.

In Service, you'll have access to clouder.node, hence all connection credentials. You'll also have access to children service and you'll not required to have Many2one fields to know them.

Application is referential data, Service is the operational one. I suspect you often confuse that concept.

Last thing, I'm not sure about delegation here. We'll want to display cert fields in clouder.service object, in a cert tab I think.

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+
+    base_id = fields.Many2one(
```
YannickB: I don't understand the use of this field. base_id name shall be reserved for many2one to clouder.base IMO

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+        string='Certificate Authority',
+        comodel_name='clouder.application',
+        required=True,
+        ondelete='cascade',
+        domain=lambda s: "[('tag_ids', '=', %d)]" % s.enf.ref(
+            'clouder_certificate_authority.tag_cert_authority',
+        ),
+    )
+    api_port_id = fields.Many2one(
```
YannickB: We shall not need this field if we use clouder.service. If you want easy access to the value, use computed field or property (https://github.com/clouder-community/clouder/blob/0.9.0/clouder/container.py#L620)

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+        string='API Port',
+        comodel_name='clouder.image.port',
+        compute='_compute_api_port_id',
+    )
+    host_id = fields.Many2one(
```
YannickB: We shall not need this field, information shall be access through node_id field

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+    certificate_request_id = fields.Many2one(
+        string='CSR',
+        comodel_name='clouder.certificate.request',
+        context="""{'default_host_ids': [(4, host_id)],
+                    'default_authority_id': id,
+                    'default_public_key_id': certificate_id,
+                    }""",
+        help='Certificate Signing Request.',
+    )
+    exec_service_id = fields.Many2one(
```
YannickB: We shall not need this field if we use clouder.service, we'll find it through child_ids field. If you want easy access to the value, use computed field or property (https://github.com/clouder-community/clouder/blob/0.9.0/clouder/container.py#L620)

clouder_certificate_authority/models/certificate_authority.py
```diff
@@ -0,0 +1,198 @@
...
+    config_id = fields.Many2one(
+        string='Configuration',
+        comodel_name='clouder.config.certificate.server',
+        help='Certificate Authority Configuration.',
+    )
+    is_initialized = fields.Boolean(
+        string='Initialized',
+        help='Has this CA server been initialized yet?',
+    )
+    openssl_service_id = fields.Many2one(
```
YannickB: We shall not need this field if we use clouder.service, we'll find it through child_ids field. If you want easy access to the value, use computed field or property (https://github.com/clouder-community/clouder/blob/0.9.0/clouder/container.py#L620)

clouder_certificate_authority/models/certificate_authority.py
YannickB: My bad, it was already a computed field. We shall decide if we prefer computed fields or property in such case.

clouder_certificate_authority/models/certificate_host.py
```diff
@@ -0,0 +1,35 @@
...
+    _description = 'Clouder Certificate Host'
+
+    name = fields.Char(
+        required=True,
+    )
+    host = fields.Char(
+        required=True,
+    )
+    port = fields.Integer()
+    api_object = fields.Binary(
```
YannickB: I don't know if we shall keep using a computed field or use a property like anywhere else in Clouder (like https://github.com/clouder-community/clouder/blob/0.9.0/clouder/container.py#L620). 
Whatever we chose, we shall use only one of them in all Clouder sourcecode, since the purpose seems to be exactly the same.

clouder_certificate_authority/api.py
YannickB: See my comment on clouder.service, it shall resolve the problem.

clouder_certificate_authority/models/certificate_host.py
lasley: Computed fields > properties because they are RecordSet aware. Properties only work with singletons, and should only be used if describing attributes of a model or data collection vs. actually being data.

YannickB: Ok I agree, then I'll support all use of computed fields in PR in the future. We'll also think about replacing properties with computed fields later.

That said, I wish computed fields would be easier to read in the code, I'd like we keep function near the field definition. That's the main added value of properties IMO

lasley: > I wish computed fields would be easier to read in the code

I agree. I've found that the OCA guidelines provide a healthy balance for this though - the computation methods should just be `_compute_my_field_name`, and should be declared in field declaration order. It at least makes it all predictable, even when in downstream modules.

clouder_certificate_authority/data/application_type.xml
```diff
@@ -0,0 +1,27 @@
...
+            model="clouder.application.type"
+            >
+        <field name="name">cfssl</field>
+        <field name="system_user">root</field>
+        <field name="tag_ids"
+               eval="[(6, 0, [ref('tag_cert_authority')])]"
+               />
+    </record>
+    
+    <record id="application_type_openssl"
```
lasley: Hmmm but they serve different purposes. What does an application type actually grant us anyways?

clouder_certificate_authority/images/openssl-exec/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+RUN apk add --no-cache openssl \
+                       openssl-dev \
+                       pip
+
+RUN pip install cryptography
+
+# Enter
+ENTRYPOINT ["openssl"]
```
lasley: Is this just because we use `execute` in order to work on the container?

clouder_certificate_authority/models/certificate_authority.py
lasley: @YannickB - Yup I totally confuse the concepts. I was honestly getting the hang of it I think, then we went and renamed everything and got me all confused again 😕 . I think I'm getting the hang though.

> Last thing, I'm not sure about delegation here. We'll want to display cert fields in clouder.service object, in a cert tab I think.

The delegation is only for a Certificate Authority, which has a bunch more keys than the standard service will. I also didn't want to litter the namespace of the service object with the stuff related to CA activities.

Are there some specific fields that you saw here that should instead be moved into the service? Or maybe some you would like elaboration on? Looking at my help strings, it seems they could use some love.

clouder_certificate_authority/data/application_type.xml
YannickB: The application type is usually used in the .py files to know which function shall be executed.

Ex: https://github.com/clouder-community/clouder/blob/0.9.0/clouder_template_odoo/template.py#L51

Until you have a template.py file in your module, you'll not really know if you need an openssl application type, but I'm willing to bet you'll not need it.

clouder_certificate_authority/images/openssl-exec/Dockerfile
YannickB: Yes, the container will be kept alive with the `cat` CMD

clouder_certificate_authority/models/certificate_authority.py
YannickB: Well the reason I'd like to avoid delegation is because I don't really see how the interface would be. Maybe with the xml view records I'll better understand the way you want the user to see the fields and send you in the best direction.

Right now you create some new models but without the corresponding views I have trouble to understand how you'll like them to be displayed.

lasley: Yeah I'll make the interface soon. I'm not really sure how we would integrate all the functionality that we need for a CA directly into any of the existing interfaces without creating clutter, so I was just going to add a "Certificate Authorities" section to provide the context. 

The goal, at least in my head, is that the CA service is managed separately from the Certificate Authorities themselves. When in the context of a CA, you will be generating and revoking keys for other services. When in the context of the service, you would manage the Certificate Authority instance such as the volumes and all that.

Given the above, this would create an isolation of the application resource (service) and the application itself (certificate authority). I think maybe it'll make sense once I do the views, so I'll try to know out in next few days.

lasley: Also keep in mind that CFSSL allows for multi-root CA, so this would sometimes require multiple CAs per one CA service.

PR #183, opened 2016-12-14
[FIX] Update Clouder Base

lasley:
Per https://github.com/clouder-community/clouder/pull/180#discussion_r92349199
* Update Clouder base in all modules


YannickB:
Thanks @lasley :)

PR #184, opened 2016-12-14
[IMP] Add Odoo 10 template

YannickB:
Let's continue our move to v10 :). Moreover, I want to base the docker compose file on v10 so I need the image for it.


clouder_template_odoo/images/odoo10-files/http.patch
```diff
@@ -0,0 +1,10 @@
+--- http.py.old 2014-10-22 11:10:40.414067190 +0200
++++ http.py     2014-10-22 11:10:52.806421122 +0200
+@@ -1379,6 +1379,7 @@
+     d, _, r = h.partition('.')
+     if d == "www" and r:
+         d = r.partition('.')[0]
++    h = h.replace('.','_').replace('-','_')
```
lasley: I'm curious - what is this for?

clouder_template_odoo/template.xml
```diff
@@ -166,6 +166,15 @@
             <field name="dockerfile"/>
         </record>
 
+        <record id="image_odoo10_files" model="clouder.image">
+            <field name="name">img_odoo_files10</field>
```
lasley: IMO we should kind of unify the id and the name here. Note the 10 is in different spots, totally going to screw with my memory/dyslexia.

clouder_template_odoo/template.xml
```diff
@@ -254,6 +272,19 @@
             <field name="update_bases" eval="True"/>
         </record>
 
+        <record id="application_odoo10_files" model="clouder.application">
+            <field name="name">Odoo 9 Files</field>
```
lasley: Odoo 10

clouder_template_odoo/template.xml
```diff
@@ -368,6 +423,22 @@
             <field name="value">clouder</field>
         </record>
 
+        <record id="application_odoo10_clouder" model="clouder.application">
+            <field name="name">Clouder 10</field>
+            <field name="code">clouder10</field>
+            <field name="type_id" ref="application_type_odoo"/>
+            <field name="template_ids" eval="[(4, [ref('clouder.application_template_base_www')])]"/>
+            <field name="child_ids" eval="[(4, [ref('application_odoo_data'),ref('application_odoo10_clouder_files'),ref('application_odoo_clouder_exec'),ref('application_odoo_ssh')])]"/>
```
lasley: While it does seem this works (I was quite boggled to learn that), the syntax is invalid. 

The ORM method `4` signature accepts one `id`, not a list of `ids`. `6` is the one that accepts a list, but also overrides all the tags that are already there.

clouder_template_odoo/images/odoo10-files/http.patch
YannickB: Oh, yeah this require explanations. I may appreciate some help here because it's an issue since the beginning of Clouder, and removing this patch will require a little lobbying to Odoo SA.

Clouder name the databases with `${application_code}_${base_domain.replace('.','_').replace('-','_')`. 
So in short we use the domain as the database name but we replace all . and - by _.

Exemple (with clouder the application_code) : 
domain : yolo-consulting.net
Database name : clouder_yolo_consulting_net
db_filter : ^clouder_%h$

Thanks to this patch, the odoo dbfilter is able to recognize the database and correctly redirect even with domain having . and - in their name. We can do multi-base odoo in clouder thanks to this patch.

Currently Odoo dbfilter don't manage dot and dash, it would be really great that this patch is integrated directly in Odoo core so we can remove it from Clouder.

clouder_template_odoo/template.xml
YannickB: Totally agree, I though I already did that.

clouder_template_odoo/template.xml
YannickB: Hell I have to agree, but the problem is in all template. I wrote it in my todo list, let's modify them all at once.

clouder_template_odoo/images/odoo10-files/http.patch
lasley: A note with name replacements - we could accidentally create duplications such as:
* hello_my.name.com
* hello-my_name.com
* hello.my.name.com

YannickB: Well postgres does not accept dash not dot in database name from what I know. I'm not sure I really got you here :/

lasley: This doesn't seem to be the case:

```
                                                   List of databases
                     Name                     |  Owner   | Encoding |   Collate   |    Ctype    |   Access privileges   
----------------------------------------------+----------+----------+-------------+-------------+-----------------------
 my-badass.domain.with-amazing_characters.com | odoo     | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
```

And the magic worked from Odoo after the `sed`:

image

Works with a database filter of `^%h$` to route automatically by domain.

YannickB: Seriously?!? What the fuck... If I went to this direction I had good reasons, and it worked as far as 9.3... ok...

Ok now that's a decision... the base fullname in Clouder has enforced underscore and is used everywhere, I can't foresee all the consequences... Thanks, I need to think about it.

lasley: 😆  an excellent response! Admittedly I didn't figure this out on light terms, much core code was scoured 🚀

YannickB: If postgres is able to manage dot and dash we'll never get the patch accepted, this is Clouder which need to be corrected. In core :(

Maybe it way because of mysql I did it this way. Anyway, I do believe we will remove the http.patch. Added on my todo list

PR #187, opened 2016-12-15
[ADD] clouder_template_red_october: File encyption/decryption

lasley:
This is a WIP, I have a question that I'll ask inline. ReadMe below for info on this template.

Clouder Template - Red October
==============================

This module provides a Clouder Template for Red Octover.

Red October is a cryptographically-secure implementation of the two-person rule
to protect sensitive data. From a technical perspective, Red October is a
software-based encryption and decryption server. The server can be used to
encrypt a payload in such a way that no one individual can decrypt it. The
encryption of the payload is cryptographically tied to the credentials of the
authorized users.

Authorized persons can delegate their credentials to the server for a period of
time. The server can decrypt any previously-encrypted payloads as long as the
appropriate number of people have delegated their credentials to the server.

This architecture allows Red October to act as a convenient decryption service.
Other systems, including CloudFlare’s build system, can use it for decryption
and users can delegate their credentials to the server via a simple web interface.
All communication with Red October is encrypted with TLS,
ensuring that passwords are not sent in the clear.

[Read More on CloudFlare's Blog](https://blog.cloudflare.com/red-october-cloudflares-open-source-implementation-of-the-two-man-rule/)

[Browse Red October on Github](https://github.com/cloudflare/redoctober>)

### DIFF
--- a/clouder_template_red_october/data/image.xml
+++ b/clouder_template_red_october/data/image.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+    <record id="image_redoctober_data" model="clouder.image">
+        <field name="name">image_redoctober_data</field>
+        <field name="template_ids"
+               eval="[(4, [ref('image_template_redoctober_data')])]"
+               />
+        <field name="parent_from">lasley/redoctober-data</field>
+    </record>
+    
+    <record id="image_redoctober_exec" model="clouder.image">
+        <field name="name">image_redoctober_exec</field>
+        <field name="template_ids"
+               eval="[(4, [ref('image_template_redoctober_exec')])]"
+               />
+        <field name="parent_from">lasley/redoctober-exec</field>
+        <field name="volumes_from">data</field>
+    </record>
+
+</odoo>
--- a/clouder_template_red_october/data/image_port.xml
+++ b/clouder_template_red_october/data/image_port.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+    
+    <record id="image_port_redoctober_http"
+            model="clouder.image.port"
+            >
+        <field name="template_id" ref="image_template_redoctober_exec" />
+        <field name="name">redoctober-http</field>
+        <field name="local_port">8888</field>
+    </record>
+
+</odoo>
--- a/clouder_template_red_october/images/data/Dockerfile
+++ b/clouder_template_red_october/images/data/Dockerfile
@@ -1,0 +1,4 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+CMD tail -f /dev/null
--- a/clouder_template_red_october/images/data/pki/ca/csr_ca.json
+++ b/clouder_template_red_october/images/data/pki/ca/csr_ca.json
@@ -1,0 +1,4 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+CMD tail -f /dev/null
--- a/clouder_template_red_october/images/exec/Dockerfile
+++ b/clouder_template_red_october/images/exec/Dockerfile
@@ -1,0 +1,37 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+
+RUN groupadd -r redoctober --gid=999 && useradd -r -g redoctober --uid=999 redoctober
+
+# Install Build Dependencies
+
+ENV buildDeps "build-base \
+               gcc \
+               git \
+               go \
+               libtool \
+               openssl \
+               runit"
+
+RUN apk add --no-cache $buildDeps
+
+# Install Red October
+
+RUN git clone --depth=1 https://github.com/cloudflare/redoctober.git /go/src/github.com/cloudflare/redoctober
+
+RUN go install github.com/cloudflare/redoctober
+
+EXPOSE 8080 8081
+ENV RO_CERTS=/var/lib/redoctober/data/server.crt \
+    RO_KEYS=/var/lib/redoctober/data/server.pem \
+    RO_DATA=/var/lib/redoctober/data \
+    RO_CERTPASSWD=password \
+    RO_COMMONNAME=localhost
+
+ENTRYPOINT ["/go/src/github.com/cloudflare/redoctober/scripts/docker-entrypoint.sh"]
+CMD ["redoctober", \
+    "-addr=:8080", \
+    "-vaultpath=/var/lib/redoctober/data/diskrecord.json", \
+    "-certs=/var/lib/redoctober/data/server.crt", \
+    "-keys=/var/lib/redoctober/data/server.pem"]
--- a/clouder_template_red_october/README.rst
+++ b/clouder_template_red_october/README.rst
@@ -1,0 +1,72 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==============================
+Clouder Template - Red October
+==============================
+
+This module provides a Clouder Template for Red Octover.
+
+Red October is a cryptographically-secure implementation of the two-person rule
+to protect sensitive data. From a technical perspective, Red October is a
+software-based encryption and decryption server. The server can be used to
+encrypt a payload in such a way that no one individual can decrypt it. The
+encryption of the payload is cryptographically tied to the credentials of the
+authorized users.
+
+Authorized persons can delegate their credentials to the server for a period of
+time. The server can decrypt any previously-encrypted payloads as long as the
+appropriate number of people have delegated their credentials to the server.
+
+This architecture allows Red October to act as a convenient decryption service.
+Other systems, including CloudFlare’s build system, can use it for decryption
+and users can delegate their credentials to the server via a simple web interface.
+All communication with Red October is encrypted with TLS,
+ensuring that passwords are not sent in the clear.
+
+`Read More on CloudFlare's Blog
+<https://blog.cloudflare.com/red-october-cloudflares-open-source-implementation-of-the-two-man-rule/>`_.
+
+`Browse Red October on Github
+<https://github.com/cloudflare/redoctober>`_.
+
+Configuration
+=============
+
+Clouder configuration instructions are available at https://clouder.readthedocs.io/
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Create a CFSSL Service in the Clouder Control Panel
+
+Known issues / Roadmap
+======================
+
+* Add more Signature Profile options - https://github.com/cloudflare/redoctober/blob/86ecfbe5750ebf05565e4c80104d0a7919792fee/doc/cmd/redoctober.txt#L113
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/clouder-community/clouder/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Dave Lasley <dave@laslabs.com>
+
+Maintainer
+----------
+
+This module is maintained by Clouder Community.
+
+To contribute to this module, please visit https://github.com/clouder-community/clouder
--- a/clouder_template_red_october/__init__.py
+++ b/clouder_template_red_october/__init__.py
@@ -1,0 +1,3 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
--- a/clouder_template_red_october/__manifest__.py
+++ b/clouder_template_red_october/__manifest__.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Clouder Template - Red October',
+    'version': '10.0.10.0.0',
+    'category': 'Clouder',
+    'depends': [
+        'clouder',
+        'clouder_template_proxy',
+    ],
+    'author': 'LasLabs Inc.',
+    'license': 'LGPL-3',
+    'website': 'https://github.com/clouder-community/clouder',
+    'data': [
+        'data/image_template.xml',
+        'data/image.xml',
+        'data/image_port.xml',
+        'data/image_volume.xml',
+        'data/application_tag.xml',
+        'data/application_type.xml',
+        'data/application_template.xml',
+        'data/application.xml',
+    ],
+    'installable': True,
+    'application': False,
+}
--- a/clouder_template_red_october/data/application.xml
+++ b/clouder_template_red_october/data/application.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+    <record id="application_redoctober_data" model="clouder.application">
+        <field name="name">CFSSL Data</field>
+        <field name="code">data</field>
+        <field name="type_id" ref="application_type_redoctober" />
+        <field name="tag_ids" eval="[(4, ref('clouder.tag_data'))]" />
+        <field name="default_image_id" ref="image_redoctober_data" />
+        <field name="sequence">1</field>
+        <field name="required" eval="True"/>
+    </record>
+
+    <record id="application_redoctober_exec" model="clouder.application">
+        <field name="name">CFSSL Exec</field>
+        <field name="code">exec</field>
+        <field name="type_id" ref="application_type_redoctober" />
+        <field name="tag_ids" eval="[(4, ref('clouder.tag_exec'))]" />
+        <field name="default_image_id" ref="image_redoctober_exec" />
+        <field name="sequence">2</field>
+        <field name="required" eval="True"/>
+        <field name="update_strategy">auto</field>
+    </record>
+
+    <record id="application_redoctober" model="clouder.application">
+        <field name="name">CFSSL</field>
+        <field name="code">redoctober</field>
+        <field name="type_id" ref="application_type_redoctober" />
+        <field name="tag_ids" eval="[(4, ref('tag_cert_authority'))]" />
+        <field name="child_ids"
+               eval="[(4, ref('application_redoctober_data')),
+                      (4, ref('application_redoctober_exec')),
+                     ]"
+               />
+        <field name="sequence">1</field>
+        <field name="required" eval="True"/>
+    </record>
+
+</odoo>
--- a/clouder_template_red_october/data/application_tag.xml
+++ b/clouder_template_red_october/data/application_tag.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+    <record id="tag_cert_authority" model="clouder.application.tag">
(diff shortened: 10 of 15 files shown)


clouder_template_red_october/data/image_port.xml
```diff
@@ -0,0 +1,15 @@
...
+<!-- Copyright 2016 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+    
+    <record id="image_port_redoctober_http"
+            model="clouder.image.port"
+            >
+        <field name="template_id" ref="image_template_redoctober_exec" />
+        <field name="name">redoctober-http</field>
```
YannickB: If you need proxy, you shall use http tag

clouder_template_red_october/images/data/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+CMD tail -f /dev/null
```
YannickB: No CMD needed, use base

clouder_template_red_october/images/data/pki/ca/csr_ca.json
```diff
@@ -0,0 +1,4 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+CMD tail -f /dev/null
```
YannickB: No CMD needed, use base

clouder_template_red_october/images/exec/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apk add --no-cache $buildDeps
+
+# Install Red October
+
+RUN git clone --depth=1 https://github.com/cloudflare/redoctober.git /go/src/github.com/cloudflare/redoctober
+
+RUN go install github.com/cloudflare/redoctober
+
+EXPOSE 8080 8081
+ENV RO_CERTS=/var/lib/redoctober/data/server.crt \
```
YannickB: I usually replace them in deploy_post hook https://github.com/clouder-community/clouder/blob/0.9.0/clouder_template_odoo/template.py#L49

I suggest you use this method for now. It'd be great to avoid using deploy_post for this purpose so we can make generic docker images which can be used without Clouder, but that's not the point of this PR.
I'm working on it, I need to make such generic image for the Clouder docker compose so let's wait until I figure the best way to do it.

In the case you describe, I still believe most of the operation will stay in deploy_post.

YannickB:
Just to be sure, redoctober is about file/data secure storage am I right ? If yes, we had a discussion about https://www.vaultproject.io/ which looked very good to me.

Can you describe what is the exact purpose for redoctober ? Shall it replace https://www.vaultproject.io/ and if yes can you explain why we shall use redoctober instead ?

lasley:
> Don't you think this template shall be merged with the cfssl template ? With your proposition we'll have

Red October and CFSSL server different purposes. Red October is file encryption, and CFSSL is a certificate authority.

CFSSL is required to secure our internal communications, such as Logstash=>Elastic.

Red October is going to handle encryption/decryption of private keys generated by CFSSL, Openssl, OpenSSH, etc.

OpenSSL container is required in order to get some certificate information that I was otherwise unable to obtain from the other services. 

As I understand it, Docker is built upon the premise of modular design & the combining of logical services is not the way to go. Technically we could just install OpenSSL on every container that requires it, but I feel like that will just increase the deploy time exponentially vs. the one container.

> Can you describe what is the exact purpose for redoctober ? Shall it replace https://www.vaultproject.io/ and if yes can you explain why we shall use redoctober instead ?

I included some of this our email thread before I noticed the question here. Just so it's all public too, here's the breakdown from the email. Let me know if it doesn't clarify.

Vaultproject duplicates a lot of functionality that the CFSSL CA does, without actually providing a CA. It does technically offer more than Red October though, specifically the storage of the encrypted data vs simply key management.

On the flip side, Red October is more secure. It has another layer of decryption rights called delegation, which allows you to delegate decryption rights to different RO Vaults. This means that the RO Vault itself is actually portable, and able to be distributed amongst multiple Vaults. In our context, this means we can add superusers to decrypt the data & template those users across customer vaults.

TBH both are about as state of the art in terms of design, for the most part it’s just the question of feature duplication. I think a lot of Vault project’s appeal (at least for me) is their website. I still need to do a side-by-side comparison though, and I’m somewhat thinking of making a Vaultproject connector simply for the hell of it.

YannickB:
Thanks @lasley for clarification, looks like the way to go for me !

clouder_template_red_october/images/exec/Dockerfile
lasley: A template language like Jinja2 would be seriously badass here if that's similar to what you're thinking. We could even use the pre-existing `odoo.tools.safe_eval` to allow us a safe code context that we control, allowing for some nice possibilities.

lasley:
I'm planning on doing a few tests with this later today & will report back on whether we can merge.

clouder_template_red_october/data/image.xml
```diff
@@ -0,0 +1,24 @@
...
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+    <record id="image_redoctober_data" model="clouder.image">
+        <field name="name">image_redoctober_data</field>
+        <field name="template_ids"
+               eval="[(4, [ref('image_template_redoctober_data')])]"
+               />
+        <field name="parent_from">clouder/base:3.4</field>
```
lasley: After inheriting from base, I am getting a build failed due to no command. Looking at the base, this makes sense (because there isn't).

Do we have another Docker layer with just a `cat` or `tail` in it that we can use instead?

YannickB: Ouch, ok I was sure we had a CMD in the inherited base image.

I made a direct commit to resolve it, can you try again with the latest image from Docker hub ? https://github.com/clouder-community/clouder/commit/8695902acd71ec8bc969b87e0a60bf58791994ea https://hub.docker.com/r/clouder/base/builds/bzb6vgezh9o8qk55ycqhedc/

lasley: Woot looks like we're in business thanks! Few more tweaks here and we should be good, I'll let you know

YannickB: As a sidenote, I shall remove all `tail -f` I made in the data images, they are all replaced by the `CMD cat` I added in the base image

clouder_template_red_october/images/exec/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM clouder/base:3.4
+MAINTAINER Dave Lasley <dave@laslabs.com>
+
+ARG RO_CERTPASSWD="password"
+ARG RO_COMMONNAME="localhost"
```
lasley: Do we have a way to specify build args programmatically? I'd like to make this common-name something more meaninful

clouder_template_red_october/data/image_port.xml
```diff
@@ -0,0 +1,23 @@
...
+    
+    <record id="image_port_redoctober_http"
+            model="clouder.image.port"
+            >
+        <field name="template_id" ref="image_template_redoctober_exec" />
+        <field name="name">https</field>
+        <field name="local_port">8080</field>
+    </record>
+
+    <record id="image_port_redoctober_metrics"
```
lasley: When deploying this from clouder, I'm getting an error when binding to 8081. This isn't happening when I build and run the image on its own, so it has to be something I'm doing wrong in the config here or there's something I don't know about on that port. Any clues where to start?

```
2017/01/02 19:52:25 core.init success: path=/var/lib/redoctober/data/diskrecord.json
2017/01/02 19:52:25 metrics.init start: addr=localhost:8081
2017/01/02 19:52:25 http.serve start: addr=:8080
2017/01/02 19:52:25 listen tcp 198.23.230.10:8081: bind: cannot assign requested address
```

Without Clouder:
```
root@odoo-dev-10:/media/sf_Repos/clouder/clouder_template_red_october/images/exec# docker run -t lasley/redoctober-exec
Generating RSA private key, 2048 bit long modulus
....+++
....................................................................................................................+++
e is 65537 (0x10001)
writing RSA key
Signature ok
subject=/C=US/ST=California/L=Everywhere/CN=localhost
Getting Private key

Generated default certificates for RedOctobeer at /var/lib/redoctober/data/server.crt and /var/lib/redoctober/data/server.pem

2017/01/02 19:56:24 core.init success: path=/var/lib/redoctober/data/diskrecord.json
2017/01/02 19:56:24 metrics.init start: addr=localhost:8081
2017/01/02 19:56:24 http.serve start: addr=:8080
```

clouder_template_red_october/images/exec/Dockerfile
YannickB: You should use environment for this. In Clouder deployment, you can specify them by overriding an hook, like this example in mysql template : 

https://github.com/clouder-community/clouder/blob/master/clouder_template_mysql/template.py#L53

Note that the use case seems to be the same, you probably want to store them in Clouder as options, like the root password of mysql.

clouder_template_red_october/data/image_port.xml
YannickB: It seems the problem is located inside your container.

`listen tcp 198.23.230.10:8081` I suspect this is not the port which can't be allocated, it's the IP adress. Probably 198.23.230.10 can be used by the node but not by the container, which has probably another IP like 172.17.*.* more often used by Docker. You probably have a setting somewhere which should be set to 0.0.0.0.

lasley: Woot good call! No idea what that IP is either, it doesn't match with any of my infrastructure (both internal and external facing). 

```
2d241b1a6c52        dev-red_october-exec-20170102.203532              "/usr/lib/go/src/gith"   24 seconds ago      Up 18 seconds                   0.0.0.0:30012->8080/tcp, 0.0.0.0:30011->8081/tcp                                         dev-red_october-exec
```

lasley:
Alright this works!

I'm not sure if the proxy part is though. Am I supposed to do something more than just setting the name as `https`?

YannickB:
So redoctober has a web interface... That's really good to know.

You'll need to add link to proxy/dns with https://github.com/clouder-community/clouder/blob/0.9.0/clouder_template_odoo/template.xml#L315, and allow base creation in the application. Then try to create a base on this service, with the url you want to access it.

lasley:
> So redoctober has a web interface... That's really good to know.

Yeah totally - I'm using the JSON API, but it comes with a rudimentary interface based on the JSON API. Makes it really freaking easy to poke around

```
$ docker run -d lasley/redoctober-exec -p 8080:8080
$ curl -k https://localhost:8080
<!DOCTYPE html>
<html lang="en">
<head>
	<title>Red October - Two Man Rule File Encryption &amp; Decryption</title>
	<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">

	<link rel="stylesheet" href="//netdna.bootstrapcdn.com/bootstrap/3.0.2/css/bootstrap.min.css" />
	<link rel="stylesheet" href="//netdna.bootstrapcdn.com/bootstrap/3.0.2/css/bootstrap-theme.min.css" />
	<script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.0.3/jquery.min.js"></script>
	<script src="//netdna.bootstrapcdn.com/bootstrap/3.0.2/js/bootstrap.min.js"></script>
	<style type="text/css">
		.footer{ border-top: 1px solid #ccc; margin-top: 50px; padding: 20px 0;}
	</style>
</head>
```

lasley:
@YannickB Hmm ok so I don't think I've tried to create a Base manually before. I created #193 because the view was basically broken, but now that I have things entered I'm getting stuck on some dependencies.

I think I have everything figured out except for the `application.default_image_id.version_ids:` constraint. What image is supposed to be linked to the application? The exec?

YannickB:
@lasley Application with children shall not have default_image_id, I'm wondering if I didn't let a wrong constraint behind. Can you provide a link to this constraint ?

lasley:
@YannickB - https://github.com/clouder-community/clouder/blob/c6d231b43cbc7e7033c3f383cad40e7c6d6aad1d/clouder/base.py#L609

YannickB:
Oh ok. This function is used when you didn't first created a service. This way you can directly create a base, it will create the service behind for the customer. Useful when you want to create a service each time you create a base for a customer, so he have his own service (no multi-base mode)

TBH, recently I only followed the process first create manually the service, then create the base and fill the service_id field so this function is not called. You should do the same for now until we debug it.

lasley:
Ok I see, so there's just a missing deploy step if you create it the one direction. I already had a service, so that was a pretty easy fix.

Hmmm I must still be missing something though, now I'm getting a missing port error. I can replicate it by Reloading the proxy, which jives in the traceback:

```
Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 638, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/http.py", line 675, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo/http.py", line 331, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo/service/model.py", line 119, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo/http.py", line 324, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo/http.py", line 933, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo/http.py", line 504, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 862, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/main.py", line 854, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 679, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 664, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/media/sf_Repos/clouder/clouder/models/base.py", line 587, in create
    return super(ClouderBase, self).create(vals)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 432, in create
    res.do('create', 'deploy_frame')
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 332, in do
    getattr(self, 'do_exec')(action, job_id)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 346, in do_exec
    getattr(self, action)()
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 360, in deploy_frame
    self.deploy_links()
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 394, in deploy_links
    link.deploy_()
  File "/media/sf_Repos/clouder/clouder/models/base_link.py", line 86, in deploy_
    'deploy_link ' + self.name.name, 'deploy_exec', where=self.base_id)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 332, in do
    getattr(self, 'do_exec')(action, job_id)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 346, in do_exec
    getattr(self, action)()
  File "/media/sf_Repos/clouder/clouder/models/base_link.py", line 93, in deploy_exec
    self.control() and self.deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_odoo/template.py", line 465, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_gitlab/template.py", line 456, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_piwik/template.py", line 130, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_shinken/template.py", line 250, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 192, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_mail/template.py", line 214, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_dns/common.py", line 64, in deploy_link
    self.base_id.generate_cert_exec()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 110, in generate_cert_exec
    proxy_link.deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_odoo/template.py", line 465, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_gitlab/template.py", line 456, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_piwik/template.py", line 130, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_shinken/template.py", line 250, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 263, in deploy_link
    port +
UnboundLocalError: local variable 'port' referenced before assignment
```

I think everything is having this trouble though, here's Gitlab:

```
2017-01-03 23:10:40,677 31878 INFO clouder odoo.addons.clouder.models.model:   File "/media/sf_Repos/clouder/clouder/models/model.py", line 360, in deploy_frame
    self.deploy_links()
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 394, in deploy_links
    link.deploy_()
  File "/media/sf_Repos/clouder/clouder/models/base_link.py", line 86, in deploy_
    'deploy_link ' + self.name.name, 'deploy_exec', where=self.base_id)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 332, in do
    getattr(self, 'do_exec')(action, job_id)
  File "/media/sf_Repos/clouder/clouder/models/model.py", line 346, in do_exec
    getattr(self, action)()
  File "/media/sf_Repos/clouder/clouder/models/base_link.py", line 93, in deploy_exec
    self.control() and self.deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_odoo/template.py", line 465, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_gitlab/template.py", line 456, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_piwik/template.py", line 130, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_shinken/template.py", line 250, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 192, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_mail/template.py", line 214, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_dns/common.py", line 64, in deploy_link
    self.base_id.generate_cert_exec()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 110, in generate_cert_exec
    proxy_link.deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_odoo/template.py", line 465, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_gitlab/template.py", line 456, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_piwik/template.py", line 130, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_shinken/template.py", line 250, in deploy_link
    super(ClouderBaseLink, self).deploy_link()
  File "/media/sf_Repos/clouder/clouder_template_proxy/template.py", line 263, in deploy_link
    port +
```

And looking back on this, I think we maybe are missing some proxy stuff in Elasticsearch. Oh well, we'll find out soon enough 😆

YannickB:
Hum I guess we're not entering this condition : 

https://github.com/clouder-community/clouder/blob/master/clouder_template_proxy/template.py#L253

Can you output `self.base_id.service_id.ports` to see where is the problem ? [IMP] : I guess we shall raise an issue here when neither http nor https are found in `self.base_id.service_id.ports`

lasley:
`Service with NAME: dev-redoctober, PORTS: {}`

lasley:
And this makes sense because the service doesn't seem to have any exposed ports:

image

YannickB:
That's strange since you specified the port in 

https://github.com/clouder-community/clouder/pull/187/files#diff-067c1eb14f6f8af45482f279b63141a4R7

But I see what the problem is. You have to expose it, either local or internet in order for the port to be inherited in service.

Eg. https://github.com/clouder-community/clouder/blob/master/clouder_template_odoo/template.xml#L122

lasley:
Damn still no go. It looks like Let's Encrypt is failing on the proxy before this (my dev doesn't have a public port 80). Think it's something to do with that? Should I maybe try a Base for something else?

YannickB:
Hum, to be honest the local expose status wasn't much tested until now, can you try with internet exposed ?

The local shall be used only for container to container interactions, so it'll not publish port in the container itself.

PR #193, opened 2017-01-02
[IMP] clouder: Upgrade base view

lasley:
* Upgrade base view from v7
* Add logical groupings and names

Looks much better IMO, although not perfect. I couldn't really think of good names/strings for some of the groups, so please suggest better ones if you have them. The top `Meta` is a good example. Maybe `Environment` instead

image

image


clouder/views/view.xml
```diff
@@ -59,77 +59,160 @@
...
+                            <newline />
+                            <button name="reset_base"
+                                    string="Reset from parent"
+                                    type="object"
+                                    attrs="{'invisible': [('reset_id','=',False)]}"
+                                    />
+                            <newline />
+                        </group>
+                        <group name="group_actions_children" string="Children">
+                            <button name="compute_childs"
```
lasley: @YannickB What do you think about moving all these buttons into the more menu? Seems like the interface would be cleaner, and this would flow with Odoo workflow a bit more I think

clouder/views/view.xml
```diff
@@ -59,77 +59,160 @@
...
+                    </group>
+                    <group name="group_build_options" string="Build Options">
+                        <group>
+                            <field name="build" />
+                            <field name="lang" />
+                            <field name="ssl_only" />
+                            <field name="public" />
+                        </group>
+                        <group>
+                            <field name="reset_id" />
```
YannickB: This shall be in a reset section

clouder/views/view.xml
```diff
@@ -59,77 +59,160 @@
...
+                                            }"
+                                    />
+                            <newline />
+                            <button name="renew_cert"
+                                    string="Renewal Certs"
+                                    type="object"
+                                    attrs="{'invisible': [('cert_renewal_date','=',False)]}"
+                                    />
+                            <newline />
+                            <button name="reset_base"
```
YannickB: This shall be in a reset section, near reset_id field

clouder/views/view.xml
YannickB: I agree, thoses would be great in the header I believe.

clouder/views/view.xml
lasley: This should be disabled if `Reset Every Day` is not selected right?

YannickB: Nope, but `Reset every day` shall be disabled is reset_id is empty. Same for button Reset from parent by the way.

In reset_id you specify the base which replace the current base. The reset action will destroy the current base, save the target base and restore it in the current base.

Eg. : 
-Base production (reset_id False, reset every day : False)

-Base formation (reset_id : production, reset every day : False). This base is used for formation purpose, to have real data. There is no risk using it and you can destroy it anytime. The reset is manually done by clicking on reset from parent button.

-Base demo (reset_id : production, reset every day : True). This base display an Odoo demo with public credential on the internet. Each day, the base is reset with the data from production base to erase any random data done by visitors during the day.

clouder/views/view.xml
lasley: Are you good with in the header instead? I moved all the buttons there other than the ones in sub-trees.

YannickB: Yes I guess it's ok

YannickB:
Ok let's merge!

PR #202, opened 2017-01-13
[IMP] Add bash in postgres image fix #201

### LINKED ISSUE #201: Postgres image does not include bash
I am testing master branch and for example gitlab template is trying to execute creation scripts using bash that is missing in postgres container.

I have installed it locally but it is a temporal workaround.

YannickB:


YannickB:
#201

YannickB:
Yep, I thought the keyword was fix :)

PR #203, opened 2017-01-14
[REF] Replace import odoo by try catch

YannickB:
Replacing all import odoo by try catch and import openerp otherwise.

I know this is gonna be a bother and will not make everyone happy, but it's too soon for me to drop support for Odoo8/Odoo9. Still, I agree that when we release Clouder 0.10 we should only support Odoo10.


lasley:
This saddens me - changes like this make repos gigantic 😦 . I guess maybe we can squash when we bring into OCA or something?

This looks fine code-wise if you feel it's necessary. I'm known for leaving older clients in the dust with no care, so my opinion probably isn't worth much in this regard 😛

YannickB:
We'll keep this until we don't have any need for older version of Odoo. Ok merging!

PR #204, opened 2017-01-14
[DEBUG] Oneclick

YannickB:
So, maybe it's finally time to make Clouder works again, don't you think?..

This PR correct most of the problems in Odoo template, in images and also database name (no more need of the patch in files, as discussed with @lasley )

Bind and postfix has been rollbacked in a previous PR to use ubuntu. We'll migrate them to alpine later, first let's get back the oneclick. Bind shall works now and my test inside the container looks good but.. I have trouble with Docker Swarm (Docker 1.12.3), it doesn't want to publish the port 53 as UDP, hence I can't fully test the oneclick.

Please feel free to try on your side after the merge, my problem is without doubt a bug and maybe it doesn't happen everywhere. I'll continue my test with libcloud DNS anyway.

Also, a check has been added in swarm mode to wait for the containers to stop/start after we update the service.

require #203


clouder/clouder_runner_docker/runner.py
```diff
@@ -445,6 +448,43 @@ def hook_purge_one(self):
...
+            'grep', self.name + '.',  '|',
+            'head -n1', '|', "awk '{print $1;}'"])
+
+    @api.multi
+    def wait_for_stop(self):
+
+        # Keep here until service is deployed
+        res = 'Start'
+        i = 1
+        total = 60
```
lasley: It would be nice if this was a method arg, defaulting to 60.

clouder/clouder_runner_docker/runner.py
```diff
@@ -445,6 +448,43 @@ def hook_purge_one(self):
...
+        while res and i <= total:
+            self.log('Stop %s' % (self.name))
+            self.log('Try %s/%s' % (i, total))
+            res = self.get_pod()
+            if res:
+                time.sleep(2)
```
lasley: It would be nice if the sleep time was also a method arg, defaulting to 2

clouder/clouder_runner_docker/runner.py
```diff
@@ -445,6 +448,43 @@ def hook_purge_one(self):
...
+                i += 1
+
+    @api.multi
+    def wait_for_start(self):
+
+        # Keep here until service is deployed
+        res = ''
+        i = 1
+        total = 60
```
lasley: Same comments re the method args. I think this could be useful, and it's a pretty simple change. Added side-effect of -1 line & -1 statically defined var in each method too 🚀

clouder/models/base.py
```diff
@@ -125,11 +125,18 @@ def fulldomain(self):
         return '%s.%s' % (self.name, self.domain_id.name)
 
     @property
+    def database(self):
```
lasley: Can we call this `database_name` instead? 

I know `database` matches the `databases` closer, but IMO it's not really explicit enough. `databases` is probably too ingrained in core to make that change right now, but it'd be nice to be explicit on a new one. I dunno, maybe I'm overthinking it.

lasley:
Actually looks like my Odoo exec is in a restart loop. Honestly I can't remember if I've actually tested the Odoo instance, I think I've only gone as far as Gitlab due to the connection error. Always just kind of figured it was dead 😆 

```
df84b7b8b79c        dev-clouder9-all-clouder9-exec-20170116.021804    "/bin/sh -c '/opt/odo"   11 minutes ago      Restarting (127) 4 minutes ago 
```

clouder/models/base.py
YannickB: How about using db_name and db_names ? We already use db_user and db_password on service : https://github.com/clouder-community/clouder/blob/master/clouder/models/service.py#L187

lasley: Yeah totally, that would be great. Shorter, still explicit, and matches pre-existing better!

YannickB:
Ok it shall be ok now, waiting for review.

@lasley The odoo exec image in Docker hub will be fixed with this PR, that's why you have the error. During my tests I use the Dockerfile field in clouder.image, the image on Docker hub shall work after the merge.
I believe it's critical now that you can use the oneclick, can you open a ticket with the issue you have ? It shall be my top priority this week, along with libcloud DNS. I also take note you use Engine, I'll make some test on this mode to make sure everything is still fine here.

clouder/models/base.py
```diff
@@ -125,32 +125,32 @@ def fulldomain(self):
...
 
     @property
-    def databases_comma(self):
+    def db_names_comma(self):
         """
         Property returning all databases names used for this base,
         separated by a comma.
         """
-        return ','.join([d for k, d in self.databases.iteritems()])
+        return ','.join([d for k, d in self.db_names.iteritems()])
```
lasley: `return ','.join(self.db_names.values())`

YannickB:
Ok, waiting for travis and then I'll merge.

That's all for now, I'll continue the oneclick debug starting from wednesday

lasley:
Looks like the template build failed https://hub.docker.com/r/clouder/odoo-exec/builds/bvqak3lcwhz6zhghyiqd72v/

YannickB:
Yes, I receive a mail when a build fail so don't worry I couldn't miss it :). Shall be corrected.

PR #227, opened 2017-05-17
WIP: Bugfix: 'NoneType' object has no attribute 'get_transport'

katyukha:
Hello!

This is mostly bugfix PR, but I want also do some more refactoring of ```__execute``` method.
Can I do it in this PR? Or it is better to create separate PR?

## Changes
- partly refactored method `ClouderModel.__execute`
- `SSHEnvironment.__getattr__`: if we define getattr we do not need to
call *super* method, because __getattr__ is called only if attribute is
not found by other methods, such as: __dict__ lookup, __getattribute__,
etc.
   - [SO question](http://stackoverflow.com/questions/3278077/difference-between-getattr-vs-getattribute)
   - [Py documentation](https://docs.python.org/2/reference/datamodel.html#object.__getattr__)

## Plans
- Refactor `ClouderModel.__execute`
  - rename it to `_execute`, this allow to override it in inherited models, especially `colouder.service`
  - move code related to `ClouderService` to `clouder.service` model (this allow to avoid strange ifs like ```if self._name == 'clouder.service'```)
  - move code that parses and wraps ```cmd``` arg to separate method


clouder/ssh/environment.py
```diff
@@ -213,10 +213,6 @@ def __init__(self, *args, **kwargs):
 
     def __getattr__(self, key):
         """ Provide passthrough to paramiko Client while locking the conn """
-        try:
```
lasley: Without this, all attributes for the `SSHEnvironment` object will be inaccessible. This try/except block first attempts attributes on the current object, then escalates to paramiko. Did you run into some sort of name conflict or something?

katyukha: No name conflicts, just attempt to make code cleaner and simpler...
I have some experience with ```__getattr__``` and ```__getattribute__``` method in few of my projects.

For example [this](https://github.com/katyukha/openerp-proxy/blob/d7eab5fc9caa9f70c1d9b629b38deb2173173b14/openerp_proxy/utils.py#L231) and [this](https://gist.github.com/katyukha/4353106#file-attrdict-py-L15) do same thing, and both working.

```__getattr__``` method is called only if attribute is not found in usual way. Thus all normal attributes of ``SSHEnvironment`` object will be accessible. The ```super``` call is required if ```__getattribute__``` method is used.

Simple repl test:

```python
class MyObject(object):
    def __init__(self, obj):
        self.attr = 42
        self.obj = obj 

    def __getattr__(self, name):
        return getattr(self.obj, name)

class C2(object):
    def __init__(self):
        self.obj_attr = 81


obj = MyObject(C2())

print obj.attr      # output: 42
print obj.obj_attr  # output: 81
```

May be i have something missed?

Also i can revert this change in next commit...

